PR #4, opened 2023-02-01
SSDS-47 initial setup and readme

james-smith-methods:


google_application_credentials.json
```diff
@@ -0,0 +1,12 @@
+{
```
dgnball, line 1: Is there a way we can avoid checking a file in that has what looks like a private key in it. Like, for instance, can we include instructions on how to generate this file? I understand that it is not meant to be secure but I'm just worried we may get some heat for how this file looks?

README.md
```diff
@@ -1,32 +1,116 @@
...
+```
+
+Note that the `FIRESTORE_PROJECT_ID` and `FIRESTORE_EMULATOR_HOST` environment variables match the settings in the `docker-compose.yml` file. The other two are covered in a later subsection.
+
+### Installing and running Firestore with Docker
+
+In order to install the Firestore emulator, and assuming that you have Docker installed, run the following command:
+
+```
+docker-compouse up -d firestore
```
tidmanmike, line 62: Spelling mistake in compose

src/scratch.py
```diff
@@ -0,0 +1,23 @@
+import os
```
dgnball, line 1: This is fine but I think we should just add a little comment to say this is for testing the database connection and maybe rename the file to "database_connection_test.py" or something like that. The other thing is that I would prefer if this lived in the integration_tests folder. The idea with the `src` folder is that is it for the python modules that make up the SDS and their corresponding unit tests.

dgnball (commented):
Please can you also resolve the Git conflict and make sure any Python code is compliant with the lint checks. Thank you

README.md
```diff
@@ -1,32 +1,116 @@
...
+
+You can use another name for the virtual environment's directory if you wish but please add it to the `.gitignore` file.
+
+### Installing the dependencies
+
+Assuming that the virtual environment is activated, run the following command:
+
+```
 pip install -r requirement.txt
 ```
```
tidmanmike, line 39: Worth adding in either here or after setting up environment variables, the command to run the application in the virtual environment? "uvicorn src.app:app --reload"

docker-compose.yml
```diff
@@ -1,11 +1,20 @@
-version: '2.1'
+version: "2.1"
 
 services:
+  firestore:
```
tidmanmike, line 4: Can we prepend the service names with "sds-" so that if a developer is working on another project they don't have to worry about conflicts when same the same docker images?

README.md
james-smith-methods: Fixed, thank you.

google_application_credentials.json
james-smith-methods: ONS reached out to me this morning soon after the commit as I suspected they would. They were satisfied with my explanation. If you have further doubts then read the discussion linked to in the readme file.

It is annoying that Firestore insists on default application credentials like this, other NoSQL database do not, but there you go.

src/scratch.py
james-smith-methods: Perhaps I should have said something, at least on Slack. It will be gone with the next ticket. Can you live with it  for this one, please? I will look at the tests and getting thing everything through the pipeline in the next ticket.

google_application_credentials.json
dgnball: Ok, fair enough.

src/scratch.py
dgnball: That's fine, if the plan is to remove it in the future.

dgnball (approved):

tidmanmike (approved):

PR #5, opened 2023-02-02
SSDS-50 Running Uvicorn and completing firebase connectivity

james-smith-methods:


dgnball (changes requested):
Please can you move any files that aren't part of SDS out of the src folder like scratch.py. You have stated that you need this to get the debugger working, but I would suggest we understand this problem before merging. Please can you also change your code so it passes the lint checks.

src/app.py
```diff
@@ -1,59 +1,79 @@
...
-@app.post("/dataset_schema")
-async def publish_schema(
-    dataset_schema_id: str, survey_id: str, payload: dict = Body(...)
-):
-    """Put a schema into the database and return the schema_id."""
-    version = database.set_schema(dataset_schema_id, survey_id, payload)
-    return {"dataset_schema_id": dataset_schema_id, "version": version}
+@app.get(UNIT_DATA_PATH)
+async def get_unit_data(dataset_id: str, unit_id: str):
+    data = database.get_data(dataset_id=dataset_id, unit_id=unit_id)
```
tidmanmike, line 47: Hi James, could you please add doc-stings to any methods that don't have any. the strand will be as follows. (I haven't had a chance to chat about our ways of working, but will book something in.

"""Summary or Description of the Function
 
 Parameters:
 argument1 (int): Description of arg1
 
 Returns:
 int:Returning value
 
"""

src/database.py
```diff
@@ -1,74 +1,112 @@
...
+
     return units_collection.document(unit_id).get().to_dict()
 
 
 def set_schema(dataset_schema_id, survey_id, dataset_schema):
     dataset_schema_versions = schemas_collection.document(dataset_schema_id)
+
     if not dataset_schema_versions.get().exists:
-        dataset_schema_versions.set({"latest_version": 1, "survey_id": survey_id})
+        dataset_schema_versions.set({LATEST_VERSION: 1, SURVEY_ID: survey_id})
```
tidmanmike, line 45: With moving everything to constants, it feels as if we will the human readable element of what we are trying to achieve is attracted away, however what might be better is to have a class model which will map to these constants to give them context. Happy to discuss this further.

src/constants.py
```diff
@@ -0,0 +1,27 @@
+OK = "OK"
```
dgnball, line 1: I think the addition of this file reduces readability. Please can you undo this change.

src/paths.py
```diff
@@ -0,0 +1,11 @@
+HEALTHCHECK_PATH = "/healthcheck"
```
dgnball, line 1: I think the addition of this file reduces readability. Please can you undo this change.

PR #9, opened 2023-02-08
SDSS-29 Get Schema details

dgnball:
- Implement the API changes for getting the schema details
- Change the app unit tests to replace the database with a mock
- Add a new function to the database module for putting the schema data in the database
- Change the integration tests to check that meta-data can be set and retrieved


src/app_test.py
```diff
@@ -30,10 +56,24 @@ def test_get_dataset_schema(client):
     assert response.status_code == 200
 
 
-def test_get_dataset_schemas(client):
+def test_query_schemas(client, database):
+    get_schemas = database.get_schemas
```
tidmanmike, line 60: Hey Daniel, looking to add in a doc string for our unit tests as well, here is the example from the other team.

`    """
     What am I testing:
     AC-1.1 - The ability to submit a CI (wellformed) to the API endpoint, and the correct response is returned with the version.
     AC-1.3 - When a CI is published in the response the datetime field is present with an ISO8601 value. (2023-01-24T13:56:38Z)
     """`

Could we add those in? Happy to discuss as a ways of working?

dgnball:
- Finish re-arrange of folders
- Time should be UTC format
- Comments on tests
- Remove surveys collection and replace with search for latest version
- Database unit tests should make an assertion

src/app_test.py
dgnball: Yes, done.

tidmanmike (approved):

martyncolmer (approved):

PR #11, opened 2023-02-15
Initial attempt at auto running on cloud run

dgnball:


craig-walker-methods (commented):
This will need the code pulling too wont it?

In mine I do:
---
steps:
- id: 'branch name'
  name: 'alpine'
  entrypoint: 'sh'  
  args: 
  - '-c'
  - | 
      echo "***********************"
      echo "$BRANCH_NAME"
      echo "***********************"

craig-walker-methods:
I would say this can be canned now.

PR #19, opened 2023-02-27
SDSS-56 event arc prototype

dgnball:
Create a Cloud Function that is initiated by uploading a new object to a storage bucket. The function will read the object out of storage and store it in Firebase.


martyncolmer (approved):

src/app/main.py
```diff
@@ -0,0 +1,31 @@
...
+    event_id = cloud_event["id"]
+    event_type = cloud_event["type"]
+
+    bucket_name = data["bucket"]
+    filename = data["name"]
+    metageneration = data["metageneration"]
+    time_created = data["timeCreated"]
+    updated = data["updated"]
+
+    print(f"Event ID: {event_id}")
```
tidmanmike, line 20: Remove prints

src/app/main.py
```diff
@@ -0,0 +1,31 @@
+import database
+import dataset_storage
+import functions_framework
+
+
+# Triggered by a change in a storage bucket
+@functions_framework.cloud_event
+def new_dataset(cloud_event):
+    data = cloud_event.data
```
tidmanmike, line 9: Doc Strings

src/app/main.py
dgnball: Done

src/app/main.py
dgnball: Done

tidmanmike (approved):

PR #20, opened 2023-03-03
Update schema file

dgnball:


sudeepkunhis (approved):

src/integration_tests/local_tests.py
```diff
@@ -70,6 +75,53 @@ def client(database, storage):
     yield client
 
 
+def test_dataset(client):
+    with open("../test_data/dataset.json") as f:
```
tidmanmike, line 79: Doc string missing and there is one print left in the code

martyncolmer (approved):

src/integration_tests/local_tests.py
dgnball: Done

tidmanmike (approved):

PR #22, opened 2023-03-14
SDSS-31 Get Dataset details

sudeepkunhis:
https://jira.ons.gov.uk/browse/SDSS-31

- Included 'sds_published_at' and 'total_reporting_units' fields in the dataset document collection in firestore
- Exposed the new dataset API endpoint - '/v1/dataset_metadata' with required request parameters - survey_id and period_id
- Added new database method to query the collection with both survey_id and period_id
- Provided exception response when dataset is not found for the specific request
- Added unit tests for the above new method and route


dgnball (changes requested):
Can you add an extra check to the test_dataset integration test. You can add it to the end. So, just call your API and loop through the results, asserting that the meta-data is what you expect.

sudeepkunhis:
@dgnball Just added the integration test as a new test, changed period_id to 'abc' in the test 'dataset.json' file since the already existing datasets in the collection don't have the new fields. Also, deployed the cloud function in the ons-sds-sandbox project, hope this looks okay now.

src/integration_tests/integration_tests.py
```diff
@@ -54,6 +55,38 @@ def test_dataset(client, bucket_loader):
     }
 
 
+def test_dataset_metadata(client, bucket_loader):
```
dgnball, line 58: So I would prefer you integrate with the existing dataset test. I've already done some work to get the sleep period right although you can increase this from 3 seconds to 5 in the conftest file. For the main integration test it doesn't need the sleep

dgnball (changes requested):

src/integration_tests/integration_tests.py
sudeepkunhis: @dgnball Got the point now, just cleaned up the tests, removed the sleep timer, the new check is now included in the same test, hope this looks okay.

dgnball (approved):

tidmanmike (approved):

src/app/database.py
```diff
@@ -13,6 +13,9 @@
 
 def set_dataset(dataset_id, dataset):
     data = dataset.pop("data")
```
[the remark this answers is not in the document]
tidmanmike: Thanks!

PR #27, opened 2023-03-19
Sds-73 local docker v1

tidmanmike:
Provides dockerized local environment for the schema and dataset endpoints. 
- Manual and http endpoint to publish datasets to the bucket.
- HTTP trigger endpoint to simulate the event arc
- thundercloud http collection for testing endpoints.


dgnball:
The Python code doesn't pass the lint tests. Can you add some extra lines to the Github action to make sure they pass the same checks that everything in src does other than unit test.

devtools/dataset-publisher/manual_insert/publish_dataset.py
dgnball, line 1: Can you add some comments to explain what this file does

README.md
```diff
@@ -3,7 +3,31 @@
 More information on this service can be found on Confluence:
 
 * https://confluence.ons.gov.uk/display/SDC/SDS
+---
+## Dockerized
+This will launch the SDS application and two emulators, the SDS application will support hot reloading within the `/src/app` directory.
```
dgnball, line 8: What is "This"? I think there are 5 containers, not 3? The meaning is not clear.

src/app/dataset_storage.py
```diff
@@ -1,9 +1,18 @@
 import json
+import os
 
 from google.cloud import storage
 
 storage_client = storage.Client()
 
+# If the local development is used then, it will use those values to create the local storage.
+if os.environ.get("DATASET_BUCKET_NAME"):
```
dgnball, line 9: You're calling `os.environ.get("DATASET_BUCKET_NAME")` three times. Can you do this once and put into a variable.

tidmanmike:
> The Python code doesn't pass the lint tests. Can you add some extra lines to the Github action to make sure they pass the same checks that everything in src does other than unit test.

I have updated the workflow file and the cloudbuild to reflect these checks.

dgnball:
I don't understand why there are two containers for emulating cloud functions when there is only one cloud function we need to emulate?

README.md
```diff
@@ -3,7 +3,31 @@
...
+The docker-compose will launch the SDS application, two storgae emulators(firebase and bucket) and two cloud functions. The SDS application will also support hot reloading within the `/src/app` directory.
 
+- You will need to create a new file called `mock_google_app_key.json` within the `/devtools` directory and copy the contents from this fake service acount found here [Mock service account](https://github.com/firebase/firebase-admin-python/blob/master/tests/data/service_account.json).
+
+This will launch the SDS application and two emulators, the SDS application will support hot reloading within the `/src/app` directory. Start the containers:
+
+```
+docker-compose up
+```
+Once loaded you can then utilise the following support tools.
```
dgnball, line 17: I would re-write this slightly to say "Once, loaded you can do the following:", because "support tools" doesn't sound right to me. Then each bullet point would start with what you can do (in bold if you like), for example:

* View the OpenAPI spec docs
* See the files put into cloud storage
* Visualise the contents of the database

... I don't really understand what "thundercloud" is?

dgnball (changes requested):

src/app/dataset_storage.py
```diff
@@ -1,9 +1,18 @@
 import json
+import os
 
 from google.cloud import storage
 
 storage_client = storage.Client()
 
+# If the local development is used then, it will use those values to create the local storage.
```
dgnball, line 8: I don't think you want to do this here because the whole point is to create the bucket object which gets overwritten by the get_dataset function anyway. You could move this whole code block to somewhere else, possible inside your devtools directory.

dgnball (changes requested):

README.md
```diff
@@ -3,7 +3,31 @@
 More information on this service can be found on Confluence:
 
 * https://confluence.ons.gov.uk/display/SDC/SDS
+---
+## Dockerized
+The docker-compose will launch the SDS application, two storgae emulators(firebase and bucket) and two cloud functions. The SDS application will also support hot reloading within the `/src/app` directory.
 
+- You will need to create a new file called `mock_google_app_key.json` within the `/devtools` directory and copy the contents from this fake service acount found here [Mock service account](https://github.com/firebase/firebase-admin-python/blob/master/tests/data/service_account.json).
```
dgnball, line 10: Are you sure you need `mock_google_app_key.json`. Last time I managed to ditch this file by setting `FIRESTORE_EMULATOR_HOST`

README.md
```diff
@@ -3,7 +3,31 @@
...
+
+This will launch the SDS application and two emulators, the SDS application will support hot reloading within the `/src/app` directory. Start the containers:
```
dgnball, line 12: This is a repetition of the sentence above.

dgnball (changes requested):

devtools/dataset-publisher/manual_insert/publish_dataset.py
tidmanmike: Added a readme within the folder explaining the purpose and operation 'devtools/dataset-publisher/README.md'

src/app/dataset_storage.py
tidmanmike: Sorted.

README.md
tidmanmike: re-worded all of this and included the curl command, we can go over this.

README.md
tidmanmike: You can test this by removing the GOOGLE_APPLICATION_KEY from the main api in the docker-compose. The code will complain that there are no credentials set.

README.md
tidmanmike: Reworded, please review

README.md
dgnball: If you can, please can you have another look at my comments and see if you agree with the suggested changes? If not, then we can leave it as is.

src/app/dataset_storage.py
dgnball: Please remove this section as it causes the build to break.

README.md
```diff
@@ -3,7 +3,56 @@
...
+- **Google cloud storage emulator** - You will be able to see files within the `devtools/gcp-storage-emulator/data/default-bucket` folder.
+
+- **Firestore emulator** - [localhost:4000/firestore](http://localhost:4000/firestore). (This data is held within the container and is lost when rebuilding.)
+
+- **thunderclient local http request collection** - The filename within devtools folder called `thunder-collection_Local Development - SDS-v1` can be imported if you are using VS Code with the thunderclient collection. With this collection you will be able to simulate all the steps currently within the SDS.  
+
+- **Simulate SDX publish process** - To best simulate a google cloud trigger with the docker containers, there is an endpoint that can be called. It will simulate the SDX publishe data process where it will imput a dataset into a cloud bucket and then call the `new_dataset` cloud function. This can be invoked with a dataset as follows:
+
+```
+curl -X POST localhost:3006 \
```
dgnball, line 28: The Curl command didn't work for me

dgnball (changes requested):

dgnball (approved):

sudeepkunhis (approved):

PR #41, opened 2023-04-04
Sds 84 schema metadata endpoint

AND-JAshcroft:
Flatten data structure returned from schema metadata endpoint and include guid in response from get schemas metadata endpoint

### DIFF
--- a/src/app/app.py
+++ b/src/app/app.py
@@ -4,7 +4,7 @@ import uuid
 import database
 import storage
 from fastapi import Body, FastAPI, HTTPException
-from models import Datasets, Schema, SchemaMetadata, Schemas
+from models import Datasets, Schema, SchemaMetadata, Schemas, ReturnedSchemaMetadata
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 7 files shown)


src/app/app.py
```diff
@@ -49,10 +49,10 @@ async def get_schema(survey_id: str, version: str) -> dict:
     return storage.get_schema(schema_metadata.schema_location)
 
 
-@app.get("/v1/schema_metadata", response_model=Schemas)
-async def query_schemas(survey_id: str) -> dict:
+@app.get("/v1/schema_metadata", response_model=list[ReturnedSchemaMetadata])
+async def get_schemas_metadata(survey_id: str) -> list[ReturnedSchemaMetadata]:
     """Retrieve the metadata for all the schemas that have a given survey_id."""
-    data = database.get_schemas(survey_id)
+    data = database.get_schema_metadata(survey_id)
```
JackRimmerAND, line 55: I think this function in database.py also be renamed to get_schemas_metadata to be consistent with this function and make it clear that it's grabbing a list of schema metadata

AND-JAshcroft: I think it already is, i changed the name from get_schemas to get_schema_metadata

JackRimmerAND: Ah, I mean the difference between get_schemas_metadata and get_schema_metadata, the missing 's'?

JackRimmerAND: Or am I misreading how it works?

AND-JAshcroft: Ah no i misread your comment, my bad. I think your right it should have the s as it refers to multiple schemas rather than returning a singular schemas metadata, will update

JackRimmerAND (approved):

sudeepkunhis (approved):

tidmanmike (approved):

PR #43, opened 2023-04-06
fixing versions in both requirements files

AND-JAshcroft:


JackRimmerAND (approved):

src/app/requirements.txt
```diff
@@ -1,2 +1 @@
-firebase_admin
```
tidmanmike, line 1: Just curious why this one has been removed as it is still used within the database.py files https://github.com/ONSdigital/sds/blob/SDSS-57-fix-versions/src/app/database.py

JackRimmerAND: It's still there on line 8 right?

tidmanmike: Yes of course, but if we are deploying and running the pip install against this requirements file, then that package won't be available as the firebase_admin package in the requirements.txt has been removed at the src/app level

I think there is ambiguity in these two requirements files having the same name. Can chat about in on Tuesday.

AND-JAshcroft: Ah i didnt realise the install was done against just this file for the deployment, i just saw that it was mirrored in the top level one, will add it back in

tidmanmike (approved):

PR #47, opened 2023-04-17
SDSS-64 - Validation of dataset JSON and metadata

sudeepkunhis:
https://jira.ons.gov.uk/browse/SDSS-64

### DIFF
--- a/src/app/dataset_storage.py
+++ b/src/app/dataset_storage.py
@@ -8,7 +8,10 @@ storage_client = storage.Client()
 
 
 def get_dataset(filename, bucket_name):
-    """Used by the cloud function."""
+    """
+    Used by the cloud function.
+    * Process if the file content is proper JSON syntax or else log error.
+    """
     bucket = storage_client.bucket(bucket_name)
     try:
         dataset = json.loads(bucket.blob(filename).download_as_string())
--- a/src/app/dataset_storage.py
+++ b/src/app/dataset_storage.py
@@ -15,7 +15,36 @@ def get_dataset(filename, bucket_name):
     bucket = storage_client.bucket(bucket_name)
     try:
         dataset = json.loads(bucket.blob(filename).download_as_string())
-        return dataset
+        isValid, missing_keys = validate_keys_dataset(dataset)
+        if isValid is True:
+            return dataset
+        else:
+            message = ""
+            for key in missing_keys:
+                message = message + key + ", "
+            logger.error(f"The keys {message} are missing in the JSON object")        
+            return None
     except ValueError as e:
         logger.error(f"Invalid JSON in the file {filename} - %s" % e)
         return None
+
+
+def validate_keys_dataset(dataset):
+    isValid = True
+    missing_keys = []
+
+    if "survey_id" not in dataset.keys():
+        missing_keys.append("survey_id")
+    if "period_id" not in dataset.keys():
+        missing_keys.append("period_id")
+    if "sds_schema_version" not in dataset.keys():
+        missing_keys.append("sds_schema_ version")
+    if "schema_version" not in dataset.keys():
+        missing_keys.append("schema_version")
+    if "form_type" not in dataset.keys():
+        missing_keys.append("schema_version")
+
+    if len(missing_keys) > 0:
+        isValid = False
+
+    return isValid, missing_keys
--- a/src/app/dataset_storage.py
+++ b/src/app/dataset_storage.py
@@ -22,7 +22,7 @@ def get_dataset(filename, bucket_name):
             message = ""
             for key in missing_keys:
                 message = message + key + ", "
-            logger.error(f"The keys {message} are missing in the JSON object")        
+            logger.error(f"The keys {message} are missing in the JSON object")
             return None
     except ValueError as e:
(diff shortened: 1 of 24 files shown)


src/app/main.py
```diff
@@ -23,12 +23,24 @@ def new_dataset(cloud_event):
...
-
-    dataset_id = str(uuid.uuid4())
-    database.set_dataset(dataset_id=dataset_id, filename=filename, dataset=dataset)
-
-    logger.info("Dataset uploaded successfully.")
+    """
+    Check if the filename ends with '.json'.
+    Process the contents if it is a valid '.json' file or else log the error.
+    """
+    if filename[-5:].lower() == ".json":
```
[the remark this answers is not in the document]
JackRimmerAND: I think we can extract this conditional so that we don't have nested if statements within it. Like this:

if filename[-5].lower() != '.json':
    logger.error(f"Invalid filetype received - {filename}")
    return

That will make the code below this more maintainable as there is less coupling between this if statement and the logic within

src/app/main.py
```diff
@@ -23,12 +23,24 @@ def new_dataset(cloud_event):
...
+            logger.debug(f"Dataset: {dataset}")
+            dataset_id = str(uuid.uuid4())
+            database.set_dataset(
+                dataset_id=dataset_id, filename=filename, dataset=dataset
+            )
+            logger.info("Dataset uploaded successfully.")
+        else:
+            logger.error("Invalid JSON file contents")
+    else:
+        logger.error(f"Invalid filetype received - {filename}")
```
JackRimmerAND, line 46: Please can you add tests for the different conditions happening here?
Tests for invalid json file names
Test for when dataset is none

src/app/dataset_storage.py
```diff
@@ -1,12 +1,53 @@
...
+        missing_keys.append("sds_schema_version")
+    if "schema_version" not in dataset.keys():
+        missing_keys.append("schema_version")
+    if "form_type" not in dataset.keys():
+        missing_keys.append("form_type")
+
+    if len(missing_keys) > 0:
+        isValid = False
+
+    return isValid, missing_keys
```
JackRimmerAND, line 53: Please could you write tests for the validation logic here? Ideally for each key we're validating plus the new logic added in get_dataset

src/app/main.py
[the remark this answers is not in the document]
sudeepkunhis: @JackRimmerAND I've corrected the if logic to the above, hope it looks better now. Also, about extracting the last 5 characters, I just checked and see that if the string itself is less than 5 characters, it just extracts the string itself, so no errors here.

src/app/main.py
sudeepkunhis: @JackRimmerAND Thanks a lot for this review, just added the logging tests as two new unit tests in 'new_dataset_logs_test.py' and another test in 'dataset_storage_test.py' specifically for the 'validate_keys' method.
I've added a pytest fixture in conftest.py for the unit test in dataset_storage_test, hope this looks okay now. I see that the test coverage has improved considerably and is now 100% for main.py and 90% for dataset_storage.py.

src/app/dataset_storage.py
sudeepkunhis: @JackRimmerAND Same as above - added a test in 'dataset_storage_test.py' specifically for the 'validate_keys' method.

src/unit_tests/dataset_storage_test.py
```diff
@@ -0,0 +1,27 @@
+import json
+
+
+def test_get_dataset_validate_keys(dataset_storage):
+    """
+    This method checks if the 'validate_keys' method ensures that each mandatory key is present in the dataset JSON.
+    """
+    with open("../test_data/dataset.json") as f:
```
AND-JAshcroft, line 8: Minor theres now a config variable that you could use here instead

src/app/main.py
```diff
@@ -23,12 +23,29 @@ def new_dataset(cloud_event):
...
-    dataset = dataset_storage.get_dataset(filename=filename, bucket_name=bucket_name)
-
-    logger.info("Dataset obtained successfully.")
-    logger.debug(f"Dataset: {dataset}")
-
-    dataset_id = str(uuid.uuid4())
-    database.set_dataset(dataset_id=dataset_id, filename=filename, dataset=dataset)
-
-    logger.info("Dataset uploaded successfully.")
+    """
```
AND-JAshcroft, line 26: Is this functionality already covered by a unit test? If not could one be added please

src/unit_tests/dataset_storage_test.py
sudeepkunhis: @AND-JAshcroft Thanks Joe, just corrected this.

src/app/main.py
sudeepkunhis: @AND-JAshcroft This was covered by the test_set_dataset in database_test.py, the test that was missing was for the invalid JSON in dataset_storage.py. I've just added an invalid JSON input in conftest.py and added a test in dataset_storage_test.py, now the coverage is 100% for dataset_storage.py. Hope this looks alright now, thanks.

AND-JAshcroft (approved):

PR #48, opened 2023-04-18
Sds 82 config file

AND-JAshcroft:


src/app/config.py
```diff
@@ -0,0 +1,161 @@
...
+    case "IntegrationTestingDocker":
+        config = IntegrationTestingLocalConfig()
+    case "unit":
+        config = UnitTestingConfig()
+    case "cloud-build":
+        config = CloudBuildConfig()
+    case "cloud-integration-test":
+        config = IntegrationTestingRemoteCloudConfig()
+    case "default":
+        config = Config()
```
JackRimmerAND, line 161: I'd argue you should put this into its own factory class, this file is getting pretty large as is.

src/app/config.py
```diff
@@ -0,0 +1,161 @@
...
+    STORAGE_EMULATOR_HOST: str
+
+
+class UnitTestingConfig(TestingConfig):
+    def __init__(self):
+        super().__init__()
+        self.SCHEMA_BUCKET_NAME = get_value_from_env("SCHEMA_BUCKET_NAME")
+
+    SCHEMA_BUCKET_NAME: str
+
```
JackRimmerAND, line 141: Is it worth putting all these classes into their own separate file? I don't tend to put multiple classes in one file but I can see why you've done it here.

src/integration_tests/conftest.py
```diff
@@ -18,13 +19,19 @@
...
+            ).exists():
+                storage_client.create_bucket(config.DATASET_BUCKET_NAME)
         except exceptions.Conflict:
             pass
         try:
-            storage_client.create_bucket(os.environ.get("SCHEMA_BUCKET_NAME"))
+            if not gcp_storage.Bucket(
+                storage_client, config.SCHEMA_BUCKET_NAME
+            ).exists():
+                storage_client.create_bucket(config.SCHEMA_BUCKET_NAME)
```
JackRimmerAND, line 34: These try excepts are doing very similar stuff, you could refactor the logic into a function with the config value as a parameter and call it twice

src/integration_tests/conftest.py
```diff
@@ -60,16 +67,19 @@ def post(self, endpoint, json):
 
 @pytest.fixture
 def client():
-    api_url = os.environ.get("API_URL")
-    if api_url:
-        if os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"):
+    if (
+        config.CONF == "IntegrationTestingCloud"
+        or config.CONF == "cloud-integration-test"
```
JackRimmerAND, line 72: Can CONF be either of these values? How come?

src/app/config.py
AND-JAshcroft: Usually id agree with you but each of these would literally only be a few lines in a file and theyre relatively tightly coupled so i think i prefer them all in one personally in this case

src/integration_tests/conftest.py
AND-JAshcroft: It can be either, i think this is just a case of terrible variable naming, will change

tidmanmike (approved):

JackRimmerAND (approved):

PR #51, opened 2023-04-21
Sdss 89 dataset unit data uplift

JackRimmerAND:


src/app/bucket/bucket_file_reader.py
```diff
@@ -0,0 +1,56 @@
...
+from logging_config import logging
+
+logger = logging.getLogger(__name__)
+
+
+class BucketFileReader:
+    def __init__(self):
+        self.storage_client = storage.Client()
+
+    def get_file_from_bucket(self, filename: str, bucket_name: str):
```
AND-JAshcroft, line 13: minor - couuld you type hint the return please

src/app/bucket/bucket_file_reader.py
```diff
@@ -0,0 +1,56 @@
...
+            isValid, message = self.validate_keys(dataset)
+            if isValid is True:
+                return dataset
+            else:
+                logger.error(
+                    f"The mandatory key(s) {message} is/are missing in the JSON object."
+                )
+                return None
+        except ValueError as e:
+            logger.error(f"Invalid JSON in the file {filename} - %s" % e)
```
AND-JAshcroft, line 28: very minor - could you use just one style of string fomatting here please

src/app/bucket/bucket_file_reader.py
```diff
@@ -0,0 +1,56 @@
...
+        ]
+        missing_keys = []
+        message = ""
+
+        for key in mandatory_keys:
+            if key not in dataset.keys():
+                missing_keys.append(key)
+
+        if len(missing_keys) > 0:
+            message = ", ".join(missing_keys)
```
AND-JAshcroft, line 53: This will mean that if any keys are missing the message will start with a comma, not a massive deal but could be irritating. Could just move the comma to being after the key

src/app/repositories/dataset_repository.py
```diff
@@ -0,0 +1,73 @@
...
+            firebase_admin.initialize_app()
+
+        self.db = firestore.client()
+        self.datasets_collection = self.db.collection("datasets")
+
+    def get_dataset_with_survey_id(
+        self, survey_id: str
+    ) -> Generator[DocumentSnapshot, None, None]:
+        """
+        Gets the survey id of a single dataset from firestore with a specific survey_id.
```
AND-JAshcroft, line 24: minor - is this documentation accurate?

src/app/services/dataset/dataset_processor_service.py
```diff
@@ -0,0 +1,158 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DatasetProcessorService:
+    def __init__(self) -> None:
+        self.dataset_repository = DatasetRepository()
+
+        self.dataset_writer_service = DatasetWriterService(self.dataset_repository)
+        pass
```
AND-JAshcroft, line 26: is this pass needed?

src/app/bucket/bucket_file_reader.py
JackRimmerAND: Done

src/app/bucket/bucket_file_reader.py
JackRimmerAND: Done

src/app/bucket/bucket_file_reader.py
JackRimmerAND: Done

src/app/repositories/dataset_repository.py
JackRimmerAND: Good spot, no it isn't will update

src/app/services/dataset/dataset_processor_service.py
```diff
@@ -0,0 +1,158 @@
...
+    ) -> None:
+        """
+        Processes the incoming dataset.
+
+        Parameters:
+        filename (str): the filename of the json containing the dataset data
+        new_dataset (NewDatasetWithMetadata): new dataset to be processed
+        """
+        logger.info("Processing new dataset...")
+        logger.debug(f"Dataset being processed: {new_dataset}")
```
AND-JAshcroft, line 37: Im not sure how i feel about logging out the datasets here and other places in this function. Even if there isnt security/GDPR reasons not too i just feel like it clutters the output. Could be worth changing to the id to allow us to still identify whats being processed but keep it a bit more compact

src/app/services/dataset/dataset_processor_service.py
```diff
@@ -0,0 +1,158 @@
...
+            f"Transformed unit data collection: {transformed_unit_data_collection}"
+        )
+
+        logger.info("Writing transformed unit data to repository...")
+        self.dataset_writer_service.write_transformed_unit_data_to_repository(
+            dataset_id, transformed_unit_data_collection
+        )
+        logger.info("Transformed unit data written to repository successfully.")
+
+    def _transform_new_dataset_metadata(
```
AND-JAshcroft, line 71: Could be worth renaming to something along the line of adding metadata to dataset or something

src/app/bucket/bucket_file_reader.py
```diff
@@ -0,0 +1,57 @@
...
+        ]
+        missing_keys = []
+        message = ""
+
+        for key in mandatory_keys:
+            if key not in dataset.keys():
+                missing_keys.append(key)
+
+        if len(missing_keys) > 0:
+            message = missing_keys.join(", ")
```
AND-JAshcroft, line 54: ah not quite what i was thinking sorry this'll end up with "key1key2key3, " was thinking more along the lines of a for key in missing keys loop

JackRimmerAND: Wouldn't it be "key1, key2, key3"?

JackRimmerAND: Because of the join keyword

AND-JAshcroft: very good point id been being an idiot 😂

src/app/services/dataset/dataset_processor_service.py
JackRimmerAND: Nope, fixed

src/app/services/dataset/dataset_processor_service.py
JackRimmerAND: Changed it to log dataset id instead

src/app/services/dataset/dataset_processor_service.py
JackRimmerAND: Yeah good point, I've renamed the unit data one below as well as it's also just adding metadata. Think I wanted to name this function to be generic in case business requirements changed but I think you're right it's better to be explicit

sudeepkunhis (approved):

AND-JAshcroft (approved):

PR #58, opened 2023-05-03
Sdss 91 error output

IsaiahLai111:


src/app/exception_throw.py
```diff
@@ -0,0 +1,94 @@
+from fastapi import status
+from fastapi.responses import JSONResponse
+
+
+class exception_response:
```
JackRimmerAND, line 5: You should be following class naming conventions i.e. Pascal Case, so this should be ExceptionResponse. Also I don't think this is the best name for the class either, I'd call it maybe ExceptionResponder. Also the class name and the file name should be consistent, when importing this class you should probably be doing e.g.

from exception_responder import ExceptionResponder

src/app/app.py
```diff
@@ -10,6 +12,25 @@
 app = FastAPI()
 
 
+@app.exception_handler(500)
+async def internal_exception_handler(request: Request, exc: Exception):
+    """
```
JackRimmerAND, line 17: How come you're passing 'request' and 'exc' as parameters when they aren't being used in here?

src/app/app.py
```diff
@@ -10,6 +12,25 @@
...
+    Override the global exception handler (500 internal server error) in
+    FastAPI and throw error in JSON format
+    """
+    return exception_throw.throw_500_global_exception()
+
+
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request: Request, exc: RequestValidationError):
```
JackRimmerAND, line 25: Same issue as above

src/app/app.py
```diff
@@ -61,10 +82,16 @@ async def get_schema(survey_id: str, version: str) -> dict:
     logger.info("Getting schema metadata...")
     logger.debug(f"Input data: survey_id={survey_id}, version={version}")
 
+    try:
```
JackRimmerAND, line 85: I wouldn't have this validation logic in app.py, ideally we'd abstract validation logic into a separate Validator class

src/app/app.py
```diff
@@ -73,19 +100,30 @@ async def get_schema(survey_id: str, version: str) -> dict:
...
+
     logger.info("Schema successfully retrieved.")
     logger.debug(f"Schema: {schema}")
 
     return schema
 
 
 @app.get("/v1/schema_metadata", response_model=list[ReturnedSchemaMetadata])
-async def get_schemas_metadata(survey_id: str) -> list[ReturnedSchemaMetadata]:
+async def get_schemas_metadata(survey_id: str = "") -> list[ReturnedSchemaMetadata]:
```
JackRimmerAND, line 114: How come we're setting a default value here? What happens if we remove this then on line 116 check 
`if survey_id is None`
instead?

src/app/app.py
```diff
@@ -95,12 +133,16 @@ async def get_schemas_metadata(survey_id: str) -> list[ReturnedSchemaMetadata]:
 
 @app.get("/v1/dataset_metadata", response_model=list[DatasetMetadata])
 async def get_dataset_metadata_collection(
-    survey_id: str, period_id: str
+    survey_id: str = "", period_id: str = ""
```
JackRimmerAND, line 136: Same issue as above, how come we're setting default values here?

src/app/database.py
```diff
@@ -122,6 +122,8 @@ def get_schema_metadata(survey_id, version) -> SchemaMetadata:
 
     for schema in schemas_result:
         return_metadata = schema.to_dict()
+        if len(return_metadata) == 0:
+            return None
```
JackRimmerAND, line 126: I don't like the idea of returning `None` here, this should probably pass the issue straight to the exception filter

src/app/exception_throw.py
```diff
@@ -0,0 +1,94 @@
...
+        self.message = message
+
+    def throw_er_with_json(self) -> JSONResponse:
+        return JSONResponse(
+            status_code=self.status_code,
+            content={"status": self.status, "message": self.message},
+        )
+
+
+def throw_500_global_exception() -> JSONResponse:
```
JackRimmerAND, line 22: How come you've defined an exception_response class and separate interceptor functions in the same file? I would separate these two concepts into different files, and probably wrap these throw functions inside an ExceptionInterceptor class.

JackRimmerAND (changes requested):

src/app/app.py
IsaiahLai111: Changed empty string to None

src/app/database.py
IsaiahLai111: Removed this logic

src/app/exception_throw.py
IsaiahLai111: Refactored into two files

src/app/storage.py
```diff
@@ -29,5 +29,8 @@ def store_schema(schema: Schema, schema_id):
 
 def get_schema(filename):
     """Get the SDS schema from the schema bucket using the filename provided."""
-    schema = json.loads(bucket.blob(filename).download_as_string())
-    return schema
+    try:
+        schema = json.loads(bucket.blob(filename).download_as_string())
+        return schema
+    except Exception:
+        return None
```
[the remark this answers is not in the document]
IsaiahLai111: Raised a custom exception at this point

src/app/app.py
IsaiahLai111: Changed from empty string to None

src/app/app.py
IsaiahLai111: Moved logic to a separate file and class

src/app/app.py
IsaiahLai111: Requirement by FASTapi on error handler

src/app/exception_throw.py
IsaiahLai111: Changed the name to follow the naming conventions

JackRimmerAND (approved):

AND-JAshcroft (approved):

PR #59, opened 2023-05-04
Sdss 94 tech debt database layer

JackRimmerAND:


src/app/routers/schema_router.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    logger.info("Schema metadata successfully posted.")
+    logger.debug(f"Schema metadata: {posted_schema_metadata}")
+
+    return posted_schema_metadata
+
+
+@router.get("/v1/schema")
+async def get_schema_metadata_from_bucket(
```
IsaiahLai111, line 43: I might be wrong here, but I believe that this endpoint is to retrieve the whole schema of a survey from JSON file instead of only the metadata, which the metadata only contains 5 attributes while response here contains the metadata and the actual schema in attribute properties as well. So I am curious about the changes in the return annotation and naming

src/app/services/shared/bucket_operations_service.py
```diff
@@ -0,0 +1,17 @@
...
+    def get_bucket_file_as_json(filename: str, bucket: Bucket) -> object:
+        """
+        Gets a file from a google cloud bucket with a specific filename and loads it as json.
+
+        Parameters:
+        filename (str): name of file being loaded.
+        bucket (Bucket): bucket being accessed.
+        """
+
+        return json.loads(bucket.blob(filename).download_as_string())
```
IsaiahLai111, line 17: I am thinking that it will be better to retain the try block here to intercept the exception of not able to find the associated file from bucket, or else it will trigger a 500 error instead of 404.
Although it should be abnormal case to not able to find the file when metadata record exists in FireStore, but returning only a 500 error with no specific logs may hinder the debug process when it happens

JackRimmerAND: Good point. I've added it to SchemaBucketRepository instead of here though, since this function is used by both the schema bucket and dataset bucket. If I added it here I'd need to create a new exception handler for a generic bucket query exception which I think is beyond the scope of this (already bloated) ticket.

src/app/repositories/buckets/dataset_bucket_repository.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        Queries google bucket for file with a specific name and returns it as json.
+
+        Parameters:
+        filename (str): name of file being queried.
+
+        Returns:
+        RawDatasetWithMetadata: raw dataset from the bucket file as json.
+        """
+        return BucketOperationsService.get_bucket_file_as_json(filename, self.bucket)
```
IsaiahLai111, line 24: It seems a bit strange that and additional service class has to be called to perform the file retrieval process, as normally it will be assumed that the repository layer will perform the retrieval task itself, in consistent with how the firebase repository classes works, instead of performing it at the service layer.
But as the file retrieval process should be consistent among schema and dataset, I recommend we can try using a parent class eg. BucketRepository to handle the get_bucket_file_as_json process, while both DatasetBucketRepository and SchemaBucketRepository inherit it to make parameters adjustment

src/app/repositories/buckets/schema_bucket_repository.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        blob = self.bucket.blob(filename)
+        blob.upload_from_string(
+            json.dumps(schema.dict(by_alias=True), indent=2),
+            content_type="application/json",
+        )
+
+    def get_bucket_file_as_json(self, filename: str) -> Schema:
+        """Get the SDS schema from the schema bucket using the filename provided."""
+        return BucketOperationsService.get_bucket_file_as_json(filename, self.bucket)
```
IsaiahLai111: Same as above

src/app/repositories/buckets/dataset_bucket_repository.py
JackRimmerAND: Yeah wasn't 100% happy with this part of the refactor, I like your suggestion and will implement it now.

JackRimmerAND: I've implemented your suggestion, let me know what you think

IsaiahLai111 (approved):

tidmanmike (approved):

PR #61, opened 2023-05-10
Modify model to support encryption

IsaiahLai111:


src/app/services/dataset/dataset_processor_service.py
```diff
@@ -190,3 +195,15 @@ def _create_dataset_metadata_item_with_id(
         metadata_collection_item["dataset_id"] = dataset_metadata_snapshot.id
 
         return metadata_collection_item
+
+    def _extract_ruref_from_unit_data(
```
JackRimmerAND, line 199: I would slightly rename this to _extract_rurefs_from_unit_data_collection, the name right now makes it seem like it extracts a single ruref from some unit data

src/app/services/dataset/dataset_processor_service.py
```diff
@@ -190,3 +195,15 @@ def _create_dataset_metadata_item_with_id(
         metadata_collection_item["dataset_id"] = dataset_metadata_snapshot.id
 
         return metadata_collection_item
+
+    def _extract_rurefs_from_unit_data(
+        self, raw_dataset_unit_data_collection: list[object]
+    ) -> list:
+        """
+        Extracts all rurefs from unit data to store in a separate list for indexing need
```
AND-JAshcroft, line 203: Minor - the docs for this function dont really want to specify whats going to happen to the data later on, its out of scope and could change.  Id personally stick to something on the lines of just 
"Extracts all rurefs from unit data to store in a separate list for indexing"

PR #64, opened 2023-05-15
Sdss 96 refactor integration tests

AND-JAshcroft:


src/app/config/config.py
```diff
@@ -1,28 +1,62 @@
...
+        super().__init__()
+        self.CONF = get_value_from_env("CONF")
+        self.TIME_FORMAT = TIME_FORMAT_STRING
+        self.API_URL = get_value_from_env("API_URL", "localhost")
+        self.DATASET_BUCKET_NAME = get_value_from_env(
+            "DATASET_BUCKET_NAME", "testDatasetBucket"
+        )
+        self.SCHEMA_BUCKET_NAME = get_value_from_env(
+            "SCHEMA_BUCKET_NAME", "testSchemaBucket"
+        )
```
JackRimmerAND, line 30: Very minor but you're using camel case for these test vars but should really be snake case for Python test_schema_bucket

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -0,0 +1,229 @@
...
+    else:
+        bucket = storage.Client().bucket(config.DATASET_BUCKET_NAME)
+        blob = bucket.blob(filename)
+        blob.upload_from_string(
+            json.dumps(dataset, indent=2), content_type="application/json"
+        )
+
+        wait_until_dataset_ready(
+            dataset["survey_id"], dataset["period_id"], session, headers
+        )
```
JackRimmerAND, line 110: Worth raising an error here if this doesn't work instead of returning 'None'?

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Returns:
+        None
+    """
+    while attempts != 0:
+        test_response = session.get(
+            f"{config.API_URL}/v1/dataset_metadata?survey_id={survey_id}&period_id={period_id}",
+            headers=headers,
+        )
+        print(test_response.status_code)
```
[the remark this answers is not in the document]
AND-JAshcroft: i did indeed

src/integration_tests/helpers/integration_helpers.py
AND-JAshcroft: Interestingly it already does as that method throws an exception if the dataset is not found

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -0,0 +1,229 @@
...
+        if Path.is_dir(dataset_bucket_path):
+            shutil.rmtree(dataset_bucket_path)
+    else:
+        _delete_blobs(storage.Client().get_bucket(config.DATASET_BUCKET_NAME))
+
+        _delete_blobs(storage.Client().get_bucket(config.SCHEMA_BUCKET_NAME))
+
+        _delete_collection(db.collection("datasets"))
+
+        _delete_collection(db.collection("schemas"))
```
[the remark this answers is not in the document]
AND-JAshcroft: As we chatted about earlier ive addresed the initial statement but kind of feel like these are fine the way they are

JackRimmerAND (approved):

cloudbuild.yaml
```diff
@@ -51,8 +51,8 @@ steps:
         --source=. \
         --entry-point=new_dataset \
         --trigger-event-filters="type=google.cloud.storage.object.v1.finalized" \
-        --trigger-event-filters="bucket=${_DATASET_BUCKET}" \
-        --set-env-vars="SCHEMA_BUCKET_NAME=${_SCHEMA_BUCKET_NAME},CONF=cloud-build"
+        --trigger-event-filters="bucket=ons-sds-sandbox-01-europe-west2-dataset-892a" \
```
tidmanmike, line 54: Why the change from variable to hardcoded value?

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -0,0 +1,284 @@
...
+    doc_ref.delete()
+
+
+MOCK_UNIT_RESPONSE = {
+    "schema_version": "v1.0.0",
+    "sds_schema_version": 4,
+    "survey_id": "test_survey_id",
+    "period_id": "abc",
+    "data": "<encrypted data>",
+}
```
IsaiahLai111, line 284: Is it better if we put all the data, either the data that run through the test or to validate the test, into test data?

cloudbuild.yaml
AND-JAshcroft: ah thats a mistake thanks for spotting :)

src/integration_tests/datasets/e2e_dataset_integration_test.py
```diff
@@ -0,0 +1,69 @@
...
+    MOCK_UNIT_RESPONSE,
+    cleanup,
+    create_dataset,
+    generate_headers,
+    load_json,
+    setup_session,
+)
+
+config = ConfigFactory.get_config()
+UNIT_ID = "43532"
```
IsaiahLai111, line 16: It might be better to put all the hard-coded data in testing in test data

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -0,0 +1,284 @@
...
+    Parameters:
+        filename: the filename to use for the dataset
+        dataset: the dataset to be created
+        session: a session instance for http/s connections
+        headers: the relevant headers for authentication for http/s calls
+
+    Returns:
+        int | None: status code for local function and no return for remote.
+    """
+    if config.API_URL.__contains__("local"):
```
IsaiahLai111, line 95: I will suggest to refractor this logic in distinguishing a local test from remote test into a specific function within helpers

AND-JAshcroft: Personally im not sure theres value in that, it would be a one line function that would make the flow of the helpers more complicated to follow

src/integration_tests/datasets/e2e_dataset_integration_test.py
AND-JAshcroft: Will move

src/integration_tests/helpers/integration_helpers.py
AND-JAshcroft: Will move

tidmanmike (approved):

PR #66, opened 2023-05-16
Sdss 105 get latest version schema with survey

IsaiahLai111:


src/app/routers/schema_router.py
```diff
@@ -59,11 +60,18 @@ async def get_schema_from_bucket(
...
+        bucket_schema_filename = (
+            schema_firebase_repository.get_latest_schema_metadata_bucket_filename(
+                survey_id
+            )
+        )
+    else:
+        bucket_schema_filename = (
+            schema_firebase_repository.get_schema_metadata_bucket_filename(
+                survey_id, version
+            )
```
JackRimmerAND, line 73: I think this if/else logic should belong in a service. Determining what function should be called based on if the version is populated is business logic and shouldn't be the responsibility of the web layer

src/integration_tests/integration_tests.py
```diff
@@ -96,3 +96,8 @@ def test_post_schema(client):
 
         assert response.status_code == 200
         assert response.json() == test_schema
+
+        response = client.get(f"/v1/schema?survey_id={schema['survey_id']}")
```
JackRimmerAND, line 100: Rather than reusing the response variable, I think you should give both get responses separate, more descriptive names for readability and to avoid mutating variables more than necessary. Maybe something like `set_version_schema_response` and `latest_version_schema_response`?

src/unit_tests/schema/get_schema_from_bucket_test.py
```diff
@@ -26,6 +26,32 @@ def test_get_schema_from_bucket_200_response(test_client):
...
+    mock_stream_obj.to_dict.return_value = {"schema_location": "test_location"}
+    mock_firebase_repository.schemas_collection.where().order_by().limit().stream.return_value = [
+        mock_stream_obj
+    ]
+
+    tmp_storage = SchemaBucketRepository.get_schema_file_as_json
+    SchemaBucketRepository.get_schema_file_as_json = MagicMock()
+    SchemaBucketRepository.get_schema_file_as_json.return_value = (
+        schema_test_data.test_schema_response
+    )
```
JackRimmerAND, line 45: How come you're mocking all the stream logic in here? Ideally we should just mock the repository functions to avoid this, I think it makes the tests confusing.

src/app/services/schema/schema_processor_service.py
```diff
@@ -105,3 +105,23 @@ def get_schema_metadata_collection_with_guid(
...
+                )
+            )
+        else:
+            bucket_schema_filename = (
+                self.schema_firebase_repository.get_schema_bucket_filename(
+                    survey_id, version
+                )
+            )
+
+        return bucket_schema_filename
```
JackRimmerAND, line 127: Sorry if this is a little nitpicky, but you don't really need to do variable assignment in here anymore, or the if/else block. So you can now do:
`
if version is None:
    return self.schema_firebase_repository.get_latest_schema_bucket_filename(survey_id)

return self.schema_firebase_repository.get_schema_bucket_filename(survey_id, version)
`

or use a ternary: 
`(
    return self.schema_firebase_repository.get_latest_schema_bucket_filename(survey_id) 
        if version is None 
        else self.schema_firebase_repository.get_schema_bucket_filename(survey_id, version)
)` 

This will cut down the amount of code you need to maintain, is still readable, and makes your code less rigid as you're no longer stuck with an if/else block

JackRimmerAND (approved):

src/integration_tests/integration_tests.py
AND-JAshcroft, line 1: This files left over from a previous iteration, its now broken out into a couple of files and updated. Could you remove it please

AND-JAshcroft (commented):
Generally looks good to me, just one comment for an artifact from a previosu version :)

src/integration_tests/integration_tests.py
IsaiahLai111: Removed file

AND-JAshcroft (approved):

PR #68, opened 2023-05-24
Sdss 112 update firestore init param

IsaiahLai111:
### Motivation and Context
This change is made so that a mock google credential is not longer required to run the app in local environment

### What has changed
- Project ID is added as an environment variable for both local and cloud environment. Default to mock-project-id for the local environment while getting the project id from GCP for the cloud environment
- DATASET_BUCKET_NAME is added to the new_dataset cloud function as an environment variable
- Mock google credential is removed from environment variable of the local docker environment
- Firebase_Admin is no longer used to connect firebase client
- Classes BucketLoader and FirebaseLoader are added to establish the storage client and firebase client respectively, and both clients are connected using the Project ID environment variable.
- BucketLoader encapsulate the function to get Schema Bucket and Dataset Bucket
- FirebaseLoader encapsulate the function to get Schema Collection and Dataset Collection
- Both loaders will return None in an UnitTest environment
- Firebase Repository, Bucket Repository, and Integration Tests are modified to use the new loader classes
- After this change, functions to delete file from bucket and to delete previous version datasets from FireStore must be mocked in UnitTest whenever invoked or error will be prompted due to the return of None type from loaders
- [Fixes] Changed integration-test-cloudbuild to use variables supplied in cloudbuild.yaml instead of the global variables. This is to allow the cloud integration test to capture the correct Project ID, buckets name, API_URL from each sandbox, thus can pass the PR triggered cloud integration test on any sandbox

### How to test?
- Ensure app can function normally in the docker and sandbox environment
- Ensure unit test can function normally
- Ensure integration test can function normally in the docker and sandbox environment
- Ensure the new_dataset cloud function can run properly in sandbox
- Ensure pr-lint-checks are passed


src/app/repositories/firebase/firebase_loader.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def get_datasets_collection(self):
+        if config.CONF != "unit":
+            self.datasets_collection = self.__get_db().collection("datasets")
+        return self.datasets_collection
+
+    def get_schemas_collection(self):
+        if config.CONF != "unit":
+            self.schemas_collection = self.__get_db().collection("schemas")
+        return self.schemas_collection
```
JackRimmerAND, line 26: Similarly to the BucketLoader I'd probably initialise this globally, setting the collection properties in the constructor and just have the getters called in the Schema and Firebase repositories

src/app/repositories/buckets/bucket_loader.py
```diff
@@ -0,0 +1,39 @@
...
+
+config = ConfigFactory.get_config()
+
+
+class BucketLoader:
+    def __init__(self):
+        self.schema_bucket = None
+        self.dataset_bucket = None
+
+    def __get_or_create_bucket(self, bucket, bucket_name):
```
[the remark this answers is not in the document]
IsaiahLai111: Will change

src/app/repositories/buckets/bucket_loader.py
```diff
@@ -0,0 +1,39 @@
...
+        self.schema_bucket = self.__get_or_create_bucket(
+            self.schema_bucket, bucket_name
+        )
+        return self.schema_bucket
+
+    def get_dataset_bucket(self, bucket_name):
+        self.dataset_bucket = self.__get_or_create_bucket(
+            self.dataset_bucket, bucket_name
+        )
+        return self.dataset_bucket
```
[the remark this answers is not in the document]
IsaiahLai111: Okay. The old way of getting the dataset bucket name from cloud event will be changed. The set and get functions will be separated, and the set function will be initialised at constructor.

src/app/repositories/firebase/firebase_loader.py
IsaiahLai111: Similar as above. Will change accordingly

src/app/repositories/firebase/firebase_loader.py
```diff
@@ -0,0 +1,42 @@
...
+config = ConfigFactory.get_config()
+
+
+class FirebaseLoader:
+    def __init__(self):
+        self.database = None
+        self.datasets_collection = None
+        self.schemas_collection = None
+        self._connect_db()
+        self._set_collections()
```
JackRimmerAND, line 13: Feels a bit weird that you're setting these values to None, then immediately setting them to actual values in `_connect_db` and `_set_collections`. Why not have this:

```
def __init__(self):
    self.database = self.connect_db()
    self.datasets_collection = self.set_collection('datasets')
    self.schemas_collection = self.set_collection('schemas')
```

`_set_collections` would change to `set_collection`, where you pass it the collection name as a parameter.

JackRimmerAND (approved):

AND-JAshcroft (approved):
LGTM 👍

PR #72, opened 2023-06-09
Sdss 108 schema transactional state

IsaiahLai111:
### Motivation and Context
In order to ensure data integrity, transactional state is implemented for the creation of schema metadata in DB and the storage of schema file. Any failed action will result all previous action to be rollback

### What has changed
A transaction decorator is used on a function that wrap the create schema and store schema process to implement the transactional state

### How to test?
Manual testing has been done to ensure data integrity can be achieved when process failure happens.
Post schema test is modified so that parameters being passed to the transaction function is asserted in normal case. Error message and response code are also asserted when there is exception within the transactional process


src/app/repositories/firebase/schema_firebase_repository.py
```diff
@@ -28,9 +32,43 @@ def get_latest_schema_with_survey_id(
...
+        schema (Schema): The schema being stored.
+        stored_schema_filename (str): Filename of uploaded json schema.
+        """
+
+        # A stipulation of the @firestore.transactional decorator is the first parameter HAS
+        # to be 'transaction', but since we're using classes the first parameter is always
+        # 'self'. Encapsulating the transaction within this function circumvents the issue.
+        @firestore.transactional
+        def post_schema_transaction_run(transaction: Transaction):
+            self.create_schema_in_transaction(
```
JackRimmerAND, line 57: Do we need to call a separate private function here? I think it would be just as clear to call `transaction.set()` directly inside of this function since it's just a one-liner and it's pretty clear what operation is being performed

src/app/repositories/firebase/schema_firebase_repository.py
```diff
@@ -39,7 +77,11 @@ def create_schema(
         schema_metadata (SchemaMetadata): The schema metadata being added to firestore.
         """
 
-        self.schemas_collection.document(schema_id).set(asdict(schema_metadata))
+        transaction.set(
+            firebase_loader.schemas_collection.document(schema_id),
```
JackRimmerAND, line 81: The `schemas_collection` is already a private field on the schema repository, so you don't need to grab it from the `firebase_loader`, you can just call `self.schemas_collection` now right?

IsaiahLai111: Will change. Missed that when I copy my function back from the transaction object

src/app/repositories/firebase/schema_firebase_repository.py
IsaiahLai111: I can do that. But my though is I treat this as a function wrapper that only calls other functions into it, so it looks more consistent in the create schema in db process and store schema process

JackRimmerAND: I can see that, but I think doing it this way makes it clearer that `create_schema_in_transaction` is actually a transaction, as in it's performing operations on the database. Storing schema in the bucket isn't a transaction, so it makes sense to wrap it in a function call. If you wrap both in a function at first glance the storing schema function might seem like it's a transaction operation.

If you still feel like this is the best way then I'm happy to approve it, but personally I think it's an unnecessary extra function to maintain

JackRimmerAND (approved):

src/app/repositories/firebase/schema_firebase_repository.py
IsaiahLai111: I see your point. My view is that this arrangement can be easier to maintain in terms of scalability that I can reuse the functions in other transactions or add other transactional functions into the wrapper. Not so obvious in the current situation but probably for possible future. I know it is a bit of overthinking here, but kind of like to have a structure that allows that to happen.

JackRimmerAND: Fair enough, although it's always good to bear in mind YAGNI (you ain't gonna need it) when you're trying to future proof something. It might be better to introduce a structure here when the new feature comes in that requires it, rather than trying to do it in advance and have extra code to maintain

tidmanmike (approved):

PR #85, opened 2023-07-12
Sdss 149 add form types

JackRimmerAND:


src/app/models/dataset_models.py
IsaiahLai111, line 1: I think form types is needed for unitdataset as well

JackRimmerAND: Ah yeah I added it to the data objects but forgot to update the model

JackRimmerAND: Fixed

IsaiahLai111 (approved):

tidmanmike (approved):

PR #93, opened 2023-08-03
Sdss 185 change schema properties to object

IsaiahLai111:
### Motivation and Context
This PR handle both SDSS-185 and SDSS-184
Schema supplied by Author has a new structure that SDS has to cope with. At the root level, field Properties has an object type instead of list. Other fields, eg. description, additionalProperties, required, are optional fields that do not need validation. Example is removed from schema

### What has changed
- Pydantic class Schema has been removed and replaced by Python type dict to allow flexibility in coping with optional fields. (Using Pydantic class to handle optional fields will result place a null value for the field even if it is not present, which is not idea)
- A SchemaValidatorService is created to validate mandatory fields, for both data type and value
- OpenAPI file is modified to match the change
- Devtool is modified to match the change
- Schema for testing is modified to use the new roofing tile and slate schema supplied by Author (see link)
- Add unit test to cover the changes

### How to test?
Pass unit test and integrator test
Test if schema is successfully stored and retrieved when
- more miscellaneous fields are present

### Links
https://jira.ons.gov.uk/browse/SDSS-184
https://jira.ons.gov.uk/browse/SDSS-185
Schema sample
https://github.com/ONSdigital/sds-schema-definitions/blob/main/schemas/roofing_tiles_and_slate.json


src/app/services/validators/schema_validator_service.py
```diff
@@ -0,0 +1,47 @@
+from exception.exceptions import ValidationException
+
+
+class SchemaValidatorService:
```
tidmanmike, line 4: Hey Isaiah, could we add a doc string to the class and describe why it is used over pydantic please. (Really to convey our discussions on pydantic vs base models as there is a lot of switching between the two over the last two months).  Chances are developers might miss it but think would be to document in validator for future reference.

tidmanmike (approved):

src/app/services/validators/schema_validator_service.py
IsaiahLai111: Have added doc string to the class explaining why dict is used over Pydantic model and the usage of this validator

PR #94, opened 2023-08-07
SDSS-182 Changes for identifying schema version

IsaiahLai111:
### Motivation and Context
Schema version no longer exists at the root level of schema but under properties > schema_version > const, a mechanism has to be developed to extract the schema version. Also, schema version has to be displayed at schema metadata for users

### What has changed
- Added validation to validate whether field schema_version exists as object under properties, and field const exists as string under schema_version and contain no empty value
- Added a flow to extract schema version from properties > schema_version > const, and store as part of schema metadata in FireStore
- Added schema_version as a field in the schema_metadata model
- Changed test data, unit test, integration test, openapi, and devtools to match the new requirement

### How to test?
- Pass unit tests and integration tests
- Pass pipeline

### Links
https://jira.ons.gov.uk/browse/SDSS-182


src/app/services/schema/schema_processor_service.py
```diff
@@ -114,6 +115,32 @@ def calculate_next_schema_version(self, schema: dict):
...
+        Get a child property from a nested dictionary
+
+        Paramenters:
+        nested_dict (dict): A nested dictionary that is being fetched
+        keys (list): A list of keys within the nested dictionary that lead to the child property
+        """
+        field = nested_dict
+
+        for key in keys:
+            field = field[key]
```
tidmanmike, line 140: Better to have the result variable be called something different such as "result", as it avoids any issues of the return being the nested dictionary passed into the function in any side effect in the future.

IsaiahLai111: Ok will change

AND-AmmarrHafiz (approved):

PR #97, opened 2023-08-15
Sdss 186 guid schema endpoint

IsaiahLai111:
### Motivation and Context
A new schema endpoint to use GUID as the query parameter to retrieve schema

### What has changed
- A new endpoint v2/schema using GET method is added to schema_router
- A validation process and exception handling method are added
- Add related unit tests and add endpoint to integration test
- Change openapi and devtools accordingly
- Update readme and openapi to match previous changes

### How to test?
Pass unit test and integration test

### Links
https://jira.ons.gov.uk/browse/SDSS-186

### DIFF
--- a/src/app/routers/schema_router.py
+++ b/src/app/routers/schema_router.py
@@ -24,6 +24,7 @@ async def post_schema(
     Posts the schema metadata to be processed.
 
     Parameters:
+    survey_id (str): survey_id of the schema
     schema (dict): schema to be processed in JSON format.
     schema_processor_service (SchemaProcessorService): injected processor service for processing the schema.
     """
@@ -77,9 +78,8 @@ async def get_schema_from_bucket(
         logger.error("Schema metadata not found")
         raise exceptions.ExceptionNoSchemaFound
 
-    logger.info("Bucket schema location successfully retrieved.")
     logger.debug(f"Bucket schema location: {bucket_schema_filename}")
-    logger.info("Getting schema...")
+    logger.info("Bucket schema location successfully retrieved. Getting schema...")
 
     schema = schema_bucket_repository.get_schema_file_as_json(bucket_schema_filename)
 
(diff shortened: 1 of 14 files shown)


src/app/routers/schema_router.py
```diff
@@ -74,9 +77,51 @@ async def get_schema_from_bucket(
         logger.error("Schema metadata not found")
         raise exceptions.ExceptionNoSchemaFound
 
-    logger.info("Bucket schema metadata location successfully retrieved.")
-    logger.debug(f"Bucket schema metadata location: {bucket_schema_filename}")
-    logger.info("Getting schema metadata...")
+    logger.info("Bucket schema location successfully retrieved.")
+    logger.debug(f"Bucket schema location: {bucket_schema_filename}")
+    logger.info("Getting schema...")
```
tidmanmike, line 82: Just housekeeping but it would be worth condensing the L80 and L82 Info logs into one info log, the L81 debug should still remain as is.

src/app/routers/schema_router.py
```diff
@@ -74,9 +77,51 @@ async def get_schema_from_bucket(
...
+    bucket_schema_filename = (
+        schema_processor_service.get_schema_bucket_filename_from_guid(guid)
+    )
+
+    if not bucket_schema_filename:
+        logger.error("Schema metadata not found")
+        raise exceptions.ExceptionNoSchemaFound
+
+    logger.info("Bucket schema location successfully retrieved.")
+    logger.debug(f"Bucket schema location: {bucket_schema_filename}")
```
tidmanmike, line 123: Same here with condensing the L122 and L124 Info logs into one info log, the L123 debug should still remain as is.

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -85,3 +85,11 @@ def test_schema_e2e(self):
 
             assert latest_version_schema_response.status_code == 200
             assert latest_version_schema_response.json() == test_schema
+
+            set_guid_schema_response = session.get(
```
tidmanmike, line 89: The double "f" looks quite confusing as it is being passed in as one argument, what do you think about something more along the lines of this `f"{config.API_URL}/v2/schema?guid={schema['guid']}"` ?

IsaiahLai111: That's a typo, thanks Mike. Will change

src/app/routers/schema_router.py
IsaiahLai111: Yeah, they should be combined. I will check whether I can merge some logs together

sudeepkunhis (approved):

tidmanmike (approved):

PR #102, opened 2023-08-22
Script to generate openapi

Mohammed1804:
### Motivation and Context

https://jira.ons.gov.uk/browse/SDSS-189?workflowName=SDSS%3A+Project+Management+Workflow&stepId=2
### How to test?

Make generate-spec


IsaiahLai111:
I think the main concern here is the deviation in error status code. The generator is using the default error code from fastapi, but since we have override most of them, they don't match.
https://fastapi.tiangolo.com/advanced/additional-responses/
This page shows that we can specify the responses code and message at the router so that the generator can use it. I am thinking we should be creating a class for each response so that it can be used at the router and also at the exception_interceptor.py 
Reference: https://stackoverflow.com/questions/62705219/in-pythons-fastapi-autogenerated-openapi-swagger-documentation-page-how-can-i
The rest of the differences in openapi.yaml mostly reside on the structure of schema, which due to the flexibility that we have offered user doesn't seems much of a problem

src/app/app.py
```diff
@@ -8,6 +8,11 @@
 logger = logging.getLogger(__name__)
 app = FastAPI()
 
+app.description = "Open api schema for SDS"
+# app.openapi_version = "2.0.0"
```
tidmanmike, line 12: No need for commented code

scripts/generate_openapi.py
```diff
@@ -0,0 +1,23 @@
...
+    with open(OUTPUT_PATH, "w+") as f:
+        yaml.dump(
+            get_openapi(
+                title=app.title,
+                version=app.version,
+                openapi_version=app.openapi_version,
+                description=app.description,
+                routes=app.routes,
+            ),
+            f,
```
IsaiahLai111, line 22: can add , sort_keys=False so that the yaml file will be in the original order that is more readable

Makefile
```diff
@@ -137,4 +137,7 @@ lint-fix:
 
 
 setup: requirements.txt
-	pip install -r requirements.txt
\ No newline at end of file
+	pip install -r requirements.txt
+
+generate-spec:
+	python -m scripts.generate_openapi
```
IsaiahLai111, line 143: Require the following env variables to work:
         export CONF=cloud-dev && \
	export PYTHONPATH=${PYTHONPATH} && \
	export SCHEMA_BUCKET_NAME=${PROJECT_ID}-europe-west2-schema && \
	export DATASET_BUCKET_NAME=${PROJECT_ID}-europe-west2-dataset && \
	export GOOGLE_APPLICATION_CREDENTIALS=${GOOGLE_APPLICATION_CREDENTIALS} && \
	export AUTODELETE_DATASET_BUCKET_FILE=${AUTODELETE_DATASET_BUCKET_FILE} && \
	export LOG_LEVEL=${LOG_LEVEL} && \
	export PROJECT_ID=${PROJECT_ID} && \
	export PUBLISH_SCHEMA_TOPIC_ID=${PUBLISH_SCHEMA_TOPIC_ID} && \
	export PUBLISH_DATASET_TOPIC_ID=${PUBLISH_DATASET_TOPIC_ID} && \

IsaiahLai111 (commented):
Probably need to add the response model to the router. Please refer to previous comment

IsaiahLai111:
I think the yaml package has to be added to requirement.txt as well

scripts/generate_openapi.py
IsaiahLai111, line 1: Probably an additional thing to think about. The path of the app and the path of the output file is hardcoded in this script, but in normal practice we should be providing these as arguments when running the script in Makefile to allow more flexibility (ie. no matter how the structure of the folder is changed in the future, all the changes should be done at the makefile instead of changing every script)
We can call the script by:
Eg. python -m scripts.generate_openapi src.app.app:app --out generate_openapi/openapi.yaml
which first argument is our fastapi app, the same in how uvicorn runs it, and second argument is the location of the output file
Happy to discuss it if needed

tidmanmike (approved):

src/app/routers/schema_router.py
IsaiahLai111, line 1: We need the responses for the rest of our end points in schema_router and dataset_router as well

generate_openapi/openapi.yaml
IsaiahLai111, line 1: Should put the file back into the original folder (gateway/ ). Change the path argument (--out) in the makefile will do

IsaiahLai111 (approved):

PR #113, opened 2023-09-11
Change parameter name from unit_id to identifier

Mohammed1804:
### Motivation and Context

in the future there is a single consistent naming pattern and will apply not only for business surveys but also for social surveys.
### What has changed

When making a request on the /v1/unit_data endpoint, the parameter is renamed to "identifier"

Unit and integration tests updated to reflect change

### Links

https://confluence.ons.gov.uk/display/SDC/Retrieve+the+unit+supplementary+data+from+SDS


IsaiahLai111 (approved):
Great job :)

tidmanmike (approved):

sudeepkunhis (approved):

PR #114, opened 2023-09-11
Sdss 207 env var for deleting datasets

IsaiahLai111:
### Motivation and Context
When a new dataset is uploaded, there is a chance that the survey has already gone live and previous version of dataset is being used. To avoid severing the linkage with the old dataset, the old dataset deletion process should not be carry out. An environment variable is introduced to control the skipping of deletion process when it is on, and deleting the latest previous dataset if it is off. This logic of deletion is to be reviewed later.

### What has changed
- Added an env variable RETAIN_DATASET_FIRESTORE
- Added a function to determine when to delete previous version dataset, and added docstring with card number to explain the changes
- Added a function in Document Version Service to calculate previous dataset version
- Added unit tests to check functions calling when the env variable is on and off
- Modify dataset integration test to upload 2 versions of dataset continuously, then check the behaviour and data accuracy
- Adjust the dataset-publish endpoint in local development to allow assertion of dataset filename in integration test
- Add and modify test data accordingly

### How to test?
Pass all unit tests and integration tests
Pass pipeline

### Links
https://jira.ons.gov.uk/browse/SDSS-207


src/app/services/dataset/dataset_writer_service.py
```diff
@@ -85,27 +85,29 @@ def try_publish_dataset_metadata_to_topic(
...
+                survey_id, period_id, previous_version
             )
-            logger.info("Previous versions deleted succesfully.")
+            logger.info("Previous version of dataset deleted succesfully.")
         except Exception as e:
             logger.debug(
-                f"Failed to delete previous versions of dataset with survey id {survey_id} \
-                and latest version {latest_version}, message: {e}"
+                f"Failed to delete previous version of dataset with survey id {survey_id} \
+                and version no {previous_version}, message: {e}"
```
Mohammed1804, line 109: Number

Mohammed1804 (approved):

PR #118, opened 2023-09-25
added docstrings to integration tests

Mohammed1804:
### Motivation and Context

Added more doc strings in both e2e_schema_integration_tests.py and e2e_dataset_integration_tests.py to explain why the tests are needed and how they are carried out. This would also improve readability.


sudeepkunhis (approved):

src/integration_tests/schemas/e2e_schema_integration_test.py
IsaiahLai111, line 1: Just a suggestion, would it be more readable if we also add the steps (the newly added one line comments) to the docstring as well? Like what we do for the dataset test

IsaiahLai111 (approved):

PR #127, opened 2023-10-31
using python:3.11-alpine base image for all cloudbuild steps

JonnyAdcockAND:
### Motivation and Context
We should make sure we use the same base image across the `Dockerfile` and `cloudbuild-dev.yaml` steps to make sure any vulnerabilities found are consistent across the containers we build.

### What has changed
* `cloudbuild-dev.yaml` updated to use `python:3.11-alpine` base image for each step


JonnyAdcockAND:
Don't need this now

PR #152, opened 2024-05-14
Sdss 737 dataset error

jamesbryer:
### Motivation and Context
Added publishing errors to new dataset error pubsub topic on 3 conditions:
- Dataset file extension is not '.json'
- Dataset content is not valid JSON
- Dataset is missing mandatory keys.

### What has changed

- Added validation for JSON content (previously unhandled error if file extension was '.json' but file content wasn't valid JSON)
- Added new pubsub message send function for dataset errors
- Separated try message publish function in dataset writer service so can be used for publishing metadata and errors. 

### How to test?
- Added new unit test for testing 
- Updated existing tests to match new expected errors
- Tested in ons-sds-jamesb-sandbox with unit tests, integration tests, and manual testing to ensure pubsub working as expected.

### Links
[Jira Ticket](https://jira.ons.gov.uk/browse/SDSS-737)
[IaC PR](https://github.com/ONSdigital/sds-iac/pull/122)


src/app/services/validators/dataset_validator_service.py
```diff
@@ -136,3 +183,18 @@ def _missing_keys_response(missing_keys: list[str]) -> tuple[bool, str]:
...
+        """
+        Publishes dataset error response to google pubsub topic.
+
+        Parameters:
+        error_message (str): error message to be published.
+        """
+
+        dataset_repository = DatasetFirebaseRepository()
+        dataset_writer_service = DatasetWriterService(dataset_repository)
+        dataset_writer_service.try_publish_dataset_error_to_topic(error_message)
```
IsaiahLai111, line 200: Just a thought on code refactoring. I am thinking it will be less convoluted if we just put `try_publish_dataset_error_to_topic` from `dataset_writer_service` to here given that the class is meant to handle the part that start writing dataset to database, calling both the `DatasetFireBaseRepository` and `DatasetWriterService` class in this validator seems not quite logical.

IsaiahLai111 (changes requested):
We will also need to add integration tests so that the dataset invalid messages can be validated. That will involve creating a new subscriber for the new topic and asserting the pub/sub message when invalid json file/ invalid file format/ missing mandatory keys are found.
It is one of the tech debt that SDS integration tests do not cover the unhappy path at the moment. So as a team we can discuss whether to do it as part of this card or save it for another card to handle.

devtools/dataset-publisher/app.py
```diff
@@ -29,9 +29,10 @@ async def dev_simulate_publish_dataset(request: Request, filename: str = None):
 
     # Save the dataset to the bucket
     blob = dataset_bucket.blob(filename)
+    data = await request.body()
```
JacobHarrisonONS, line 32: reasoning behind these changes?

src/integration_tests/datasets/e2e_dataset_integration_test.py
```diff
@@ -6,26 +6,37 @@
...
+    test_dataset_error_subscriber_id,
+    test_dataset_subscriber_id,
+)
 
 
 class E2EDatasetIntegrationTest(TestCase):
     def setUp(self) -> None:
         cleanup()
         pubsub_setup(dataset_pubsub_helper, test_dataset_subscriber_id)
+        pubsub_setup(dataset_error_pubsub_helper, test_dataset_error_subscriber_id)
```
IsaiahLai111, line 34: I have been running the integration tests and occasionally it will prompt an error saying that the subscriber cannot be found (404). I think it resemble an existing bug that the subscriber is not properly setup after it is teared down from the previous tests. I have tried to add a line that pause the test for 5 seconds and it seems to solve the issue, at least temporarily.
I recommend adding the line `time.sleep(5)` in front of `pubsub_setup` for both dataset and schema integration tests

devtools/dataset-publisher/app.py
```diff
@@ -29,9 +29,10 @@ async def dev_simulate_publish_dataset(request: Request, filename: str = None):
...
     blob.upload_from_string(
-        json.dumps(await request.json(), indent=2),
-        content_type="application/json",
+        data.decode("utf-8"),
+        content_type="application/octet-stream",
```
IsaiahLai111, line 35: Suggest to just omit this parameter and use the default `text/plain` content_type

src/integration_tests/datasets/e2e_dataset_integration_test.py
```diff
@@ -358,3 +372,99 @@ def test_dataset_without_title(self):
...
+        dataset_incorrect_extension_filename = create_filename_error_filepath(
+            "integration-test-incorrect-extension"
+        )
+
+        create_dataset_response = create_dataset(
+            dataset_incorrect_extension_filename,
+            dataset_incorrect_extension,
+            session,
+            headers,
+        )
```
IsaiahLai111, line 408: The `wait_until_dataset_ready` procedure will not be necessary in this case as the process deemed to fail at the validation stage. Recommend the add a flag to skip the procedure.

src/integration_tests/datasets/e2e_dataset_integration_test.py
```diff
@@ -358,3 +372,99 @@ def test_dataset_without_title(self):
...
+
+        # Upload dataset with missing keys
+        dataset_missing_keys = load_json(
+            f"{config.TEST_DATASET_PATH}dataset_missing_keys.json"
+        )
+        dataset_missing_keys_filename = create_filepath("integration-test-missing-keys")
+
+        create_dataset_response = create_dataset(
+            dataset_missing_keys_filename, dataset_missing_keys, session, headers
+        )
```
IsaiahLai111, line 456: Same as above

IsaiahLai111 (approved):

JacobHarrisonONS (approved):
Looks good

PR #153, opened 2024-05-22
Batch process

oms09:
### Motivation and Context
Transaction has been used to write dataset metadata and unit data into the FireStore, but contains a hard limit of maximum size that it cannot process large dataset (with unit data > 20k)
Transaction is replaced by Batch Write to handle large dataset. However as atomicity only exist for each batch but not across batches, if any batch write triggers an error, although should rarely happen, a clean up function is to be triggered to clean up all data residuals from FireStore. After batch write, a unit_data count checking function is used to make sure the batch write process is successful

### What has changed
- Batch write has replaced transaction in writing dataset and unit data to FireStore
- Batch delete is now used in clean up function that delete data residuals if any batch triggers an error
- The count of unit_data is checked with the value in `total_reporting_unit` as a safety net to ensure the batch write process is completed without problem
- Batch delete logic is now used to delete previous version of dataset when the `retain_dataset_firestore` flag is disabled
- A unit_data count checking function is used to make sure batch write is successful. The function count records by batch using cursor to prevent 530 query timeout error on large dataset
- Set the dataset upload function timeout to maximum of 540 seconds to process large dataset. This is the maximum timeout seconds for event triggered function, which will still timeout for larger dataset (~ 300K) and should be handled in another Jira ticket.
- Set the dataset upload function with 512 MiB memory to process large dataset

### How to test?
- Pass unit test and integration test
- Manual test the clean up function by triggering an exception is one of the batches to check if the data residuals (previously written data of the dataset) is being cleaned up successfully
- To test the previous version dataset deletion logic, change the `retain_dataset_firestore` flag value to FALSE in the gcloud command that deploy the new-dataset function, then deploy the function and upload twice of a dataset containing >20k unit data record. Check if the last version of dataset is deleted completely from FireStore.

### Links
https://jira.ons.gov.uk/browse/SDSS-809

### Issues not fixed
- 540 seconds of maximum timeout will halt the publish process of larger dataset (~300K). GCloud deploy command disallow to override this setting even though manual deploy on console can exceed this limit.


src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +37,89 @@ def get_latest_dataset_with_survey_id_and_period_id(
 
         return dataset_metadata
 
-    def perform_new_dataset_transaction(
+    def perform_batched_dataset_write(
```
JacobHarrisonONS, line 40: this function name is used in the writer service and firebase repository. could we rename one to avoid any confusion.
(perform_batched_dataset_write is called within perform_batched_dataset_write)

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +37,89 @@ def get_latest_dataset_with_survey_id_and_period_id(
...
+            batch.set(new_dataset_document, dataset_metadata_without_id, merge=True)
+            batch.commit()
+
+            batch_counter = 0
+            batch = self.client.batch()
+
+            for i in range(len(unit_data_collection_with_metadata)):
+                if batch_counter > 0 and batch_counter % self.BATCH_SIZE == 0:
+                    batch.commit()
+                    batch = self.client.batch()
```
JacobHarrisonONS, line 73: you are setting batch a second time, is this required?

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +37,89 @@ def get_latest_dataset_with_survey_id_and_period_id(
...
+
+                new_unit = unit_data_collection_snapshot.document(
+                    extracted_unit_data_identifiers[i]
+                )
+                batch.set(new_unit, unit_data_collection_with_metadata[i], merge=True)
+                batch_counter += 1
+
+            batch.commit()
```
JacobHarrisonONS, line 81: this final commit will only be required if there is a remainder on the unit_data_collection_with_metadata % 500 where it will loop and never become divisible by 500 without a remainder. is there any effect to batch.commit() when the batch is empty? I wonder if there is any point is defining the batch size by survey type or size of data because these smaller surveys are probably not going to fill a batch

IsaiahLai111: This commit is always going to be needed as the other commit, the one that run between batches, will only serve when the data is > batch_size. Either we have data count that is just divisible by 500, or we have remainder data count for the last batch, this commit will serve its purpose to commit the last batch.
We are now taking the maximum size a batch can hold, which I think is the best because atomicity is preserved within a batch and we need not to run through the clean up process if anything exception occur within a single batch. That'll save time as the deletion is running quite slow.

src/app/repositories/firebase/dataset_firebase_repository.py
IsaiahLai111: Will change the name

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +37,89 @@ def get_latest_dataset_with_survey_id_and_period_id(
...
-            for unit_data, identifier in zip(
-                unit_data_collection_with_metadata,
-                iter(extracted_unit_data_identifiers),
-            ):
-                new_unit = unit_data_collection_snapshot.document(identifier)
-                transaction.set(new_unit, unit_data, merge=True)
+            raise RuntimeError("Error performing batched dataset write.")
 
-        dataset_transaction(self.client.transaction())
+    def delete_collection_in_batches(
```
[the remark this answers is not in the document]
IsaiahLai111: Yes I will be splitting it

src/app/repositories/firebase/dataset_firebase_repository.py
IsaiahLai111: I think it is necessary to reset the batch after each commit based on this answer (although it is in Javascript, I just can't find a proper Python solution unfortunately)
Reference: https://stackoverflow.com/questions/66143646/run-multiple-firestore-batch-commit

src/app/repositories/firebase/dataset_firebase_repository.py
JacobHarrisonONS: is it worth doing if not batch_size % 500 then batch.commit commit if there was a remainder then you will need to commit after the loop? or do you not see any point?

src/app/repositories/firebase/dataset_firebase_repository.py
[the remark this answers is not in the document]
IsaiahLai111: I have split the function, one for dataset metadata and one for its subcollection. And I have changed to use batch.delete that run quicker than deleting each document sequentially (although the GCP example is using the sequential delete) as "batch delete"
Reference: https://firebase.google.com/docs/firestore/manage-data/delete-data#collections

src/app/repositories/firebase/dataset_firebase_repository.py
IsaiahLai111: Because the commit that run between batches will only be run when there is a number of record of batch_size % 500 + 1. Eg. if we have 501 records, then 1-500 batch will be committed and open another batch for the last record.
If we have exactly 500 records, this commit is still necessary to commit the batch.

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -92,6 +152,26 @@ def get_unit_supplementary_data(
...
+        Parameters:
+        dataset_id (str): The unique id of the dataset
+        """
+
+        returned_unit_data = (
+            self.datasets_collection.document(dataset_id).collection("units").stream()
+        )
+
+        unit_data_count = 0
+        for unit_data in returned_unit_data:
```
JacobHarrisonONS, line 170: unit_data is unused

IsaiahLai111: I have switched to count unit_data with cursor as the old way cause 530 query timed out error

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +38,114 @@ def get_latest_dataset_with_survey_id_and_period_id(
...
-            new_dataset_document = self.datasets_collection.document(dataset_id)
+        try:
+            doc = self.datasets_collection.document(dataset_id).get()
 
-            transaction.set(
-                new_dataset_document, dataset_metadata_without_id, merge=True
-            )
-            unit_data_collection_snapshot = new_dataset_document.collection("units")
+            logger.info("Deleting subcollection in batches")
+            for subcollection in doc.reference.collections():
```
JacobHarrisonONS, line 110: nit pick sorry rename to sub_collection following coding style

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -35,44 +38,114 @@ def get_latest_dataset_with_survey_id_and_period_id(
...
+            doc.reference.delete()
 
-        dataset_transaction(self.client.transaction())
+        except Exception as e:
+            logger.error(f"Error deleting dataset: {e}")
+            raise RuntimeError("Error deleting dataset.")
+
+    def delete_subcollection_in_batches(
+        self,
+        subcollection_ref: firestore.CollectionReference,
```
JacobHarrisonONS, line 121: same here sub_collection_ref

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -132,51 +258,16 @@ def perform_delete_previous_version_dataset_transaction(
...
-        doc_ref: the reference of the document being deleted.
-        """
-        for collection_ref in doc_ref.collections():
-            self._delete_collection(transaction, collection_ref)
+        for dataset in previous_version_dataset:
+            dataset_id = dataset.id
+            logger.info(f"dataset id {dataset_id}")
 
-        transaction.delete(doc_ref)
+        self.delete_collection_in_batches(self.datasets_collection, 100, dataset_id)
```
JacobHarrisonONS, line 272: i get an warning that dataset_id may not be set

IsaiahLai111: Thanks for picking that up. I have forgotten to change the function call after changing the name. Will fix

jamesbryer (approved):

JacobHarrisonONS (approved):

PR #154, opened 2024-06-05
Sdss 806 auto create bucket/pubsub topic removal

jamesbryer:
### Motivation and Context
SDS will now return 500 global exception if bucket / pubsub topic cannot be found. This removes the logic that will create these resources within SDS if they don't exist. This should prevent the creation of 'self-titled' buckets and removes resource creation from SDS as this is handled by Terraform. 

### What has changed
- Removed logic to autocreate buckets and pubsub topics if they don't exist
- Added exception handlers for these cases to raise 500 global exception if these resources cannot be found. 

### How to test?
- Tested using local and sandbox integration/unit tests as well as manual testing in Dockerised env to ensure endpoints return 500 error when buckets cannot be found.

### Links
[Jira ticket](https://jira.ons.gov.uk/browse/SDSS-806)

### DIFF
--- a/src/app/app.py
+++ b/src/app/app.py
@@ -77,6 +77,14 @@ app.add_exception_handler(
     exceptions.ExceptionNoSurveyIDs,
     ExceptionInterceptor.throw_404_no_survey_id_exception,
 )
+app.add_exception_handler(
+    exceptions.ExceptionBucketNotFound,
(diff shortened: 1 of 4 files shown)


IsaiahLai111 (approved):
Looks all good. Nice work

JacobHarrisonONS (approved):

PR #160, opened 2024-08-01
Sdss 712 http trigger implementation

jamesbryer:
### What has changed
- Changes from SDSS-820 (link to PR)
- Updated Cloud Function to collect oldest file in bucket for processing
- ~~Change cloudbuild YAML to run build-cloud-function script instead of deploying separately~~
- ~~Created build-cloud-function script to trigger update-new-dataset-cloud-function~~

### How to test?
- Tested pipleine on ons-sds-jb
- Unit and integration tests pass
- Tested new dataset Cloud Function to 150,000 unit data size (500,000 was attempted but Cloud Function rand out of memory)

### Links
[Jira ticket](https://jira.ons.gov.uk/browse/SDSS-712)
[SDS IaC PR](https://github.com/ONSdigital/sds-iac/pull/130)


src/app/repositories/buckets/dataset_bucket_repository.py
```diff
@@ -21,3 +21,17 @@ def get_dataset_file_as_json(self, filename: str) -> RawDataset:
...
         return self.get_bucket_file_as_json(filename)
+
+    def fetch_first_filename_from_bucket(self) -> str | None:
+        """
+        Fetches the first filename from the bucket.
+
+        Returns:
+        str: filename from the bucket.
+        """
+        blobs = self.bucket.list_blobs()
```
IsaiahLai111, line 32: Can you try to find if there is a way to retrieve the first file according to the creation time? That is to handle the 2nd issue in spike report

src/integration_tests/datasets/e2e_dataset_integration_test.py
IsaiahLai111, line 1: Can you add a function that will delete all files in dataset bucket before E2E test run? That is to handle the 4th issue described in the spike report

IsaiahLai111 (approved):
All good to me

README.md
```diff
@@ -155,19 +155,18 @@ gcloud auth login
 gcloud config set project $PROJECT_NAME
 
 cd src/app/
-gcloud functions deploy new-dataset-function \
---allow-unauthenticated \
+gcloud functions deploy http-new-dataset-function \
+--no-allow-unauthenticated \
 --gen2 \
---ingress-settings=internal-and-gclb \
+--ingress-settings=all \
```
JacobHarrisonONS, line 161: https://cloud.google.com/functions/docs/networking/network-settings
is internal only suitable rather than allowing all

cloudbuild-dev.yaml
```diff
@@ -108,18 +108,17 @@ steps:
       - |
         cd src/app/
         gcloud functions deploy new-dataset-function \
-        --allow-unauthenticated \
+        --no-allow-unauthenticated \
         --gen2 \
-        --ingress-settings=internal-and-gclb \
+        --ingress-settings=all \
```
JacobHarrisonONS, line 113: same here 
https://cloud.google.com/functions/docs/networking/network-settings

cloudbuild-prod.yaml
```diff
@@ -50,18 +50,17 @@ steps:
       - |
         cd src/app/
         gcloud functions deploy new-dataset-function \
-        --allow-unauthenticated \
+        --no-allow-unauthenticated \
         --gen2 \
-        --ingress-settings=internal-and-gclb \
+        --ingress-settings=all \
```
JacobHarrisonONS, line 55: ingress settings again

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -7,7 +7,7 @@
 
 
 class DatasetFirebaseRepository:
-    WRITE_BATCH_SIZE = 500
+    WRITE_BATCH_SIZE = 100
```
JacobHarrisonONS, line 10: any reason batch size is reduced ?

src/app/services/dataset/dataset_bucket_service.py
```diff
@@ -18,18 +18,23 @@ def get_and_validate_dataset(self, filename: str) -> RawDataset:
...
         filename: name of file being retrieved from bucket
         """
-        DatasetValidatorService.validate_file_is_json(filename)
+        isValid, message = DatasetValidatorService.validate_file_is_json(filename)
+        if not isValid:
+            if config.AUTODELETE_DATASET_BUCKET_FILE is True:
+                self.try_delete_bucket_file(filename)
 
-        raw_dataset = self.dataset_bucket_repository.get_dataset_file_as_json(filename)
+            raise RuntimeError(message)
```
JacobHarrisonONS, line 26: better to throw a more accurate error rather than run time like TypeError rather than a Runtime. i guess this wont use the custom error handler since im unsure how that would get handled with the cloud function. 
https://docs.python.org/3/library/exceptions.html#TypeError

src/app/services/dataset/dataset_bucket_service.py
```diff
@@ -18,18 +18,23 @@ def get_and_validate_dataset(self, filename: str) -> RawDataset:
         Parameters:
         filename: name of file being retrieved from bucket
         """
-        DatasetValidatorService.validate_file_is_json(filename)
+        isValid, message = DatasetValidatorService.validate_file_is_json(filename)
```
JacobHarrisonONS, line 21: rename isValid to is_valid

src/app/repositories/firebase/dataset_firebase_repository.py
IsaiahLai111: That is to avoid hitting the maximum size limit (10Mb) per write operation. We can have a method to dynamically control it depending on the unit_data size later.

src/app/services/dataset/dataset_bucket_service.py
IsaiahLai111: The monitoring module is only capturing the RuntimeError from cloud function, so that's why I have not choose to use a more specific error type.

JacobHarrisonONS (approved):
Looks good. Great work

PR #176, opened 2024-09-18
Tune config for SDS application on cloudbuild-dev and cloudbuild-prod

IsaiahLai111:
### Motivation and Context
A follow up tuning of the SDS configuration after the scale test so that SDS can achieve best performance

### What has changed
Max concurrency per SDS container is set to 1 while increasing the max-instance to 100 and 500 respectively for sandboxes and prod/preprod/integration/performance environments.
Configurations across projects are referenced here:
https://confluence.ons.gov.uk/display/SDC/SDS+-+Configurations

### How to test?
All pipelines are passed.

### Links
https://jira.ons.gov.uk/browse/SDSS-680

### DIFF
--- a/cloudbuild-dev.yaml
+++ b/cloudbuild-dev.yaml
@@ -97,7 +97,9 @@ steps:
         "--min-instances",
         "1",
         "--max-instances",
-        "2",
+        "100",
+        "--concurrency",
+        "1",
       ]
 
   - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ More information on this service can be found on Confluence:
 
 - https://confluence.ons.gov.uk/display/SDC/SDS
 
----
-
 ## Dockerized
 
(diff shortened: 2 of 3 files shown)


cloudbuild-dev.yaml
```diff
@@ -97,7 +97,9 @@ steps:
         "--min-instances",
         "1",
         "--max-instances",
-        "2",
+        "100",
+        "--concurrency",
+        "1",
```
jamesbryer, line 102: Is this needed if concurrency is 1 by default?

jamesbryer (approved):

cloudbuild-dev.yaml
JacobHarrisonONS: i guess its there for completeness

JacobHarrisonONS (approved):

cloudbuild-dev.yaml
IsaiahLai111: Technically I can omit it here as it is default by Terraform. But yeah just for completeness purpose

PR #178, opened 2024-09-25
Change integration test to only tear down only test related data

IsaiahLai111:
### Motivation and Context
- To avoid deleting FireStore data not relevant to the test
- To shorten the time required running the test

### What has changed
- Integration test teardown process will now delete FireStore data only if the survey_id matches test data pattern {test_survey_id%} (test_survey_id, test_survey_id_1, test_survey_id_abc123 etc.).
- The teardown process will also only delete schema files under survey_id that matches the test data pattern as stated above
- Change the deletion function in the teardown process to use batch delete to increase the speed
- Annotate the organise the test data. Make sure all test data survey_id conform to the test data pattern

### How to test?
- Publish schema and dataset to GCP without using the test_survey_id% as survey_id
- Run integration test on the GCP
- Make sure all test data do not remain in FireStore and buckets
- Make sure all data not related to test remain in FireStore and buckets

### Links
https://jira.ons.gov.uk/browse/SDSS-837

FireStore query simulate LIKE operator:
https://stackoverflow.com/questions/46568142/google-firestore-query-on-substring-of-a-property-value-text-search
https://stackoverflow.com/a/63460539


src/integration_tests/helpers/firestore_helpers.py
```diff
@@ -3,41 +3,77 @@
...
     """
-    doc_collection = collection_ref.stream()
+
+    # Query the collection for documents equivalent to survey_id LIKE "test_survey_id%"
+    # \uf8ff is a unicode character that is greater than any other character
+    doc_collection = (
+        collection_ref.where('survey_id', '>=', test_survey_id)
+        .where('survey_id', '<=', test_survey_id + '\uf8ff')
+        .stream()
+    )
```
IsaiahLai111, line 22: Query simulating the LIKE operator. Reference link in comment.

jamesbryer (approved):

JacobHarrisonONS (approved):

PR #179, opened 2024-09-27
Sdss 839 dynamic batch size

oms09:
### Motivation and Context
This change dynamically adjusts the batch size based on the size of unit data (up to a max of 9MB for write and delete), ensuring the system can handle datasets of various sizes without crashing. Previously, batch size was statically limited by count.

### What has changed
- Removed the static limit of 100 units per batch.
- Added logic to calculate the batch size dynamically based on the size of serialized data.
- Committed the batch early if the 9MB limit is reached.
- Reworked deletion process by iterating over documents in batches of 10, dynamically tracking document size to ensure total batch size doesn't exceed 9MB.

### How to test?
- Run the code with a dataset containing various unit sizes.
- Ensure no errors occur during large dataset processing.
- use the link on cloud function performance test to copy cpu, memory configuration and to compare its time.
- you can log it out to confirm batch size is dynamically adjusted and committed when the 9MB limit is reached
- you can force an exception in the perform_batched_dataset_write(...) method after loop to trigger and test the delete batch function. 

### Links
https://jira.ons.gov.uk/browse/SDSS-839
https://confluence.ons.gov.uk/display/SDC/Cloud+Function+Performance+Testing


src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -7,7 +9,7 @@
 
 
 class DatasetFirebaseRepository:
-    WRITE_BATCH_SIZE = 100
+    MAX_BATCH_SIZE_BYTES= 9 * 1024 * 1024
```
jamesbryer, line 12: Space between constant name and assignment operator. What is the reason for the calculation rather than the value?

jamesbryer (changes requested):

JacobHarrisonONS (changes requested):
No tests.

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -138,6 +147,21 @@ def delete_sub_collection_in_batches(
...
+        Calculate the size of serialized object in bytes.
+
+        Parameters:
+        obj (Any): The object to be serilaized and measured
+
+        Returns:
+        int: The size of the serialized object in bytes
+        """
+        serialized_obj = json.dumps(obj)
+        return len(serialized_obj.encode('utf-8'))
```
IsaiahLai111, line 162: Move this function to shared_service as it is not part of database interaction.
Also, write unit test for this function.

src/app/repositories/firebase/dataset_firebase_repository.py
IsaiahLai111, line 1: Will need the same dynamic batch logic in dataset deletion

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -7,7 +9,7 @@
 
 
 class DatasetFirebaseRepository:
-    WRITE_BATCH_SIZE = 100
+    MAX_BATCH_SIZE_BYTES= 9 * 1024 * 1024
     DELETE_BATCH_SIZE = 100
```
IsaiahLai111, line 13: As mentioned above, the deletion batch should be dynamic as well

src/app/repositories/firebase/dataset_firebase_repository.py
oms09: 1024 is the size of KB so I thought it would be more readable like this instead of putting something like 9437184

JacobHarrisonONS: (1024 bytes = 1 KB) x (1024 = 1 MB) x 10 = 10 MB

src/app/repositories/firebase/dataset_firebase_repository.py
```diff
@@ -118,26 +126,29 @@ def delete_sub_collection_in_batches(
         sub_collection_ref (firestore.CollectionReference): The reference to the sub collection
         """
         try:
-            docs = sub_collection_ref.limit(self.DELETE_BATCH_SIZE).get()
-            doc_count = 0
-
+            docs = sub_collection_ref.get()
```
IsaiahLai111, line 129: I am worried that this will use up too much memory.
To strike a balance, my thought is to obtain 10 documents at a time, and continue to fetch documents into the batch if 90% of batch size is not reached.
You can reference to the function `get_number_of_unit_supplementary_data_with_dataset_id` to how to use cursor, probably it will become `sub_collection_ref.limit(10).order_by("__name__").start_after(cursor).stream()`
Make sure you test it with large dataset and trigger an exception at publish only when you also have large amount of unit data already inserted to test the functionality
Also record the time required to delete the unit data so that you will know if optimisation is needed

IsaiahLai111 (approved):
Looks good

jamesbryer (approved):

JacobHarrisonONS (approved):

PR #180, opened 2024-09-30
Sdss 836 improve docker build efficiency

jamesbryer:
### Motivation and Context
Speed up build times and separate unit tests and linting to workflow to improve dev build efficiency

### What has changed
- Improve Dockerfile efficiency by installing requirements before copying source code to harness docker chache
- Add --cache-from flag in `docker build` command to use previous Artifact Registry image with `latest` tag as basis for new image (this is best practice for all Docker builds in Cloud Build, see documentation linked below).
- Move unit tests and linting from cloudbuild to an action so pipeline will still build for devs when testing even with linting/unit-test failures (won't risk unlinted/untested code from going to prod asthese are now an additional PR check required to pass for merge).

### Links
[Jira ticket](https://jira.ons.gov.uk/browse/SDSS-836)
[Speeding up Cloud Build doc](https://cloud.google.com/build/docs/optimize-builds/speeding-up-builds)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3.11-alpine
-COPY src/app src
-COPY requirements.txt requirements.txt
+COPY requirements.txt .
 RUN pip install -r requirements.txt
+COPY src/app src
 ENV PYTHONPATH=src
-CMD exec uvicorn src.app:app --host 0.0.0.0 --port $PORT
+CMD ["uvicorn", "src.app:app", "--host", "0.0.0.0", "--port", "8080"]
```
IsaiahLai111, line 6: `CMD ["sh", "-c", "uvicorn", "src.app:app", "--host", "0.0.0.0", "--port", "$PORT"]`
Can we execute the shell in the exec form as mentioned here https://docs.docker.com/reference/dockerfile/#shell-and-exec-form ? Might need to check if Fastapi like it.

cloudbuild-dev.yaml
```diff
@@ -17,33 +17,16 @@ steps:
...
-        make unit-test
-
   - name: docker
     id: "docker build and push"
     entrypoint: sh
     args:
       - "-c"
       - |
-        docker build -t "europe-west2-docker.pkg.dev/${PROJECT_ID}/sds/sds:${SHORT_SHA}" -t "europe-west2-docker.pkg.dev/${PROJECT_ID}/sds/sds:latest" .
+        docker build --cache-from "europe-west2-docker.pkg.dev/${PROJECT_ID}/sds/sds:latest" -t "europe-west2-docker.pkg.dev/${PROJECT_ID}/sds/sds:${SHORT_SHA}" -t "europe-west2-docker.pkg.dev/${PROJECT_ID}/sds/sds:latest" .
```
IsaiahLai111, line 26: Will it cause any issue if the SDS with :latest tag does not exist?

jamesbryer: It will continue with no cache if image isn't found, I tested this behaviour on my sandbox by deleting :latest tag from most recent image in artifact registry

Dockerfile
jamesbryer: as per docs kept uvicorn command in 1 element which worked :)

Dockerfile.develop
```diff
@@ -3,4 +3,4 @@ COPY src/app src
 COPY requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 ENV PYTHONPATH=src
-CMD exec uvicorn src.app:app --reload --host 0.0.0.0 --port $PORT
+CMD exec uvicorn src.app:app --reload --host 0.0.0.0 --port $PORT
```
IsaiahLai111, line 6: Better to align both docker file to use exec form

IsaiahLai111 (approved):

Mohammed1804:
Not sure how much of a difference it would make but separating the linting and unit test would allow the tests to run parallel/at the same time instead of sequentially which could improve the time. So it would look something like this

jobs:
  lint:
    name: Lint Python Code
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version-file: .python-version

      - name: Install dependencies
        run: make setup

      - name: Lint Python Code
        run: make lint

  test:
    name: Run Unit Tests on Python Code
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version-file: .python-version

      - name: Install dependencies
        run: make setup

      - name: Run Unit Tests
        run: make unit-test

Otherwise looks good 👍

Mohammed1804 (approved):

PR #185, opened 2024-10-22
Sdss 840 simplify int tests schema endpoint happy paths

jamesbryer:
### Motivation and Context
The pre-existing monilithic structure to schema router integration tests has been deemed innappropriate by the team. Therefore, we decided to split the tests into individual test methods for each endpoint of the router.

### What has changed
- Added tests for multiple schemas and more robust test for retrieving latest schema
- Moved each endpoints test to its own test method within the test class
- Used class level attributes to store information between tests and setup session and headers
- Removed tearDown() method to prevent test data being cleared between tests running and moved relevant functionality to teardown_class() method to be run after all test methods in the class. 
- Added test ordering via pytest-order plugin to ensure tests have the data they need before running
- Bumped fastapi due to vulnerability

### How to test?
- Run integation tests

### Links
[Jira ticket](https://jira.ons.gov.uk/browse/SDSS-840)
[PyTest Ordering docs](https://pypi.org/project/pytest-order/)

### DIFF
--- a/src/integration_tests/helpers/integration_helpers.py
+++ b/src/integration_tests/helpers/integration_helpers.py
@@ -291,18 +291,16 @@ def cleanup() -> None:
 
         client = firebase_loader.get_client()
 
-        for survey_id in test_survey_id_list:
-            perform_delete_on_collection_with_test_survey_id(
-                client,
-                firebase_loader.get_datasets_collection(),
-                survey_id
-            )
-        for survey_id in test_survey_id_list:
-            perform_delete_on_collection_with_test_survey_id(
-                client,
-                firebase_loader.get_schemas_collection(),
-                survey_id
-            )
+        perform_delete_on_collection_with_test_survey_id(
+            client,
+            firebase_loader.get_datasets_collection(),
+            test_survey_id
+        )
+        perform_delete_on_collection_with_test_survey_id(
+            client,
+            firebase_loader.get_schemas_collection(),
+            test_survey_id
+        )
 
 
 def pubsub_setup(pubsub_helper: PubSubHelper, subscriber_id: str) -> None:
--- a/src/integration_tests/helpers/integration_helpers.py
+++ b/src/integration_tests/helpers/integration_helpers.py
@@ -21,7 +21,7 @@ from src.integration_tests.helpers.firestore_helpers import (
     perform_delete_on_collection_with_test_survey_id,
 )
 from src.integration_tests.helpers.pubsub_helper import PubSubHelper
-from src.test_data.dataset_test_data import test_survey_id
+from src.test_data.dataset_test_data import test_survey_id, test_survey_id_list
 
 storage_client = storage.Client()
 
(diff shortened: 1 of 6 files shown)


IsaiahLai111 (commented):
We still need to test the situations of multiple publish of schema with the same and different survey_id. But that will constitute to scope creep so probably we should put it into another ticket.
Other than that everything looks good, only a minor comment

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -17,43 +17,38 @@
 
 
 class E2ESchemaIntegrationTest(TestCase):
+    schema_metadatas = None
```
JacobHarrisonONS, line 20: typo?

jamesbryer: schema_metadatas is the response from /v1/schema_metadata and so is a list of metadata used in the for loop within the test methods

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -290,16 +291,18 @@ def cleanup() -> None:
...
-            client,
-            firebase_loader.get_datasets_collection(),
-            test_survey_id
-        )
-        perform_delete_on_collection_with_test_survey_id(
-            client,
-            firebase_loader.get_schemas_collection(),
-            test_survey_id
-        )
+        for survey_id in test_survey_id_list:
```
JacobHarrisonONS, line 294: the delete works based on the prefix

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -12,122 +12,173 @@
...
+                assert set_version_schema_response.json() == self.test_schema
+
+                # Verify schema retrieval by the latest version
+                latest_version_schema_response = self.session.get(
+                    f"{config.API_URL}/v1/schema?survey_id={schema_metadata['survey_id']}",
+                    headers=self.headers,
+                )
+
+                assert latest_version_schema_response.status_code == 200
+                assert latest_version_schema_response.json() == self.test_schema
```
IsaiahLai111, line 129: Same here, it can't validate if it is returning the right schema for us as they are the same schema

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -12,122 +12,173 @@
...
+        for survey_id in test_survey_id_list:
+            for schema_metadata in E2ESchemaIntegrationTest.schema_metadatas_dict[survey_id]:
+                # Verify schema retrieval by GUID
+                set_guid_schema_response = self.session.get(
+                    f"{config.API_URL}/v2/schema?guid={schema_metadata['guid']}",
+                    headers=self.headers,
+                )
+
+                assert set_guid_schema_response.status_code == 200
+                assert set_guid_schema_response.json() == self.test_schema
```
IsaiahLai111, line 168: Same as previous

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -12,122 +12,181 @@
...
+        for survey_id in test_survey_id_list:
+            for index, schema_metadata in enumerate(E2ESchemaIntegrationTest.schema_metadatas_dict[survey_id]):
+                # Verify schema retrieval by GUID
+                set_guid_schema_response = self.session.get(
+                    f"{config.API_URL}/v2/schema?guid={schema_metadata['guid']}",
+                    headers=self.headers,
+                )
+
+                assert set_guid_schema_response.status_code == 200
+                assert set_guid_schema_response.json() == self.test_schemas[index]
```
IsaiahLai111, line 175: Testing the retrieval of 1 schema via GUID should suffice

IsaiahLai111 (approved):

JacobHarrisonONS (approved):

PR #189, opened 2024-11-11
SDS-847-integration-tests-unhappy path

Mohammed1804:
### Motivation and Context

Integration test not exhaustive and properly split up. Integration tests currently are lengthy and complex covering multiple endpoints. The preference is to simplify and split into tests for individual endpoints for the happy paths and separate integration tests for unhappy paths
### How to test?

Run integration test

### DIFF
--- a/src/integration_tests/schemas/e2e_schema_integration_test_unhappy_path.py.py
+++ b/src/integration_tests/schemas/e2e_schema_integration_test_unhappy_path.py.py
@@ -1,0 +1,170 @@
+from unittest import TestCase
+import pytest
+from src.app.config.config_factory import config
+from src.integration_tests.helpers.integration_helpers import (
+    cleanup,
+    generate_headers,
+    pubsub_setup,
+    pubsub_teardown,
+    setup_session,
+    inject_wait_time,
+)
+from src.integration_tests.helpers.pubsub_helper import schema_pubsub_helper
+from src.test_data.schema_test_data import test_survey_id
+from src.test_data.shared_test_data import test_schema_subscriber_id
+
+class E2ESchemaIntegrationUnhappyPaths(TestCase):
+    session = None
+    headers = None
+    invalid_survey_id = "nonexistent_survey"
+    invalid_data = {"invalid": "data"}
+    invalid_token_headers = None
+
+    @classmethod
+    def setup_class(cls) -> None:
+        cleanup()
+        pubsub_setup(schema_pubsub_helper, test_schema_subscriber_id)
+        inject_wait_time(3) 
+        cls.session = setup_session()
+        cls.headers = generate_headers()
+        cls.invalid_token_headers = {"Authorization": "Bearer invalid_token"}
+
+    @classmethod
+    def teardown_class(cls) -> None:
+        cleanup()
+        inject_wait_time(3) # Allow all messages to process.
+        pubsub_teardown(schema_pubsub_helper, test_schema_subscriber_id)
+
+    @pytest.mark.order(1)
+    def test_post_schema_unauthorized(self):
+        """
+        Test unauthorized access by providing incorrect authorization token for POST /v1/schema.
+        * Expected: 401 Unauthorized.
+        """
+        response = self.session.post(
+            f"{config.API_URL}/v1/schema?survey_id={test_survey_id}",
+            json=self.invalid_data,
+            headers=self.invalid_token_headers,
+        )
+        assert response.status_code == 401
+
+    @pytest.mark.order(2)
+    def test_post_schema_validation_error(self):
+        """
+        Test validation issue by providing invalid data for POST /v1/schema.
+        * Expected: 400 Bad Request.
+        """
+        response = self.session.post(
+            f"{config.API_URL}/v1/schema?survey_id={test_survey_id}",
+            json=self.invalid_data,
+            headers=self.headers,
+        )
+        assert response.status_code == 400
+
+    @pytest.mark.order(3)
+    def test_get_schema_404_not_found(self):
+        """
+        Test data not found by requesting a nonexistent schema ID for GET /v1/schema.
+        * Expected: 404 Not Found.
+        """
+        response = self.session.get(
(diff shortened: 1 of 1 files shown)


src/integration_tests/schemas/test_integration_unhappy_path.py
```diff
@@ -0,0 +1,170 @@
+from unittest import TestCase
+import pytest
+from src.app.config.config_factory import config
+from src.integration_tests.helpers.integration_helpers import (
+    cleanup,
+    generate_headers,
+    pubsub_setup,
+    pubsub_teardown,
+    setup_session,
+    inject_wait_time,
```
JacobHarrisonONS, line 10: this is used to clear down pubsub

src/integration_tests/schemas/test_integration_unhappy_path.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @classmethod
+    def teardown_class(cls) -> None:
+        cleanup()
+        inject_wait_time(3) # Allow all messages to process
+        pubsub_teardown(schema_pubsub_helper, test_schema_subscriber_id)
+
+    @pytest.mark.order(1)
+    def test_post_schema_unauthorized(self):
+        """
```
JacobHarrisonONS, line 40: could docstring follow layout similar to James PR
 """
        Test for /v1/dataset_metadata endpoint without passing survey_id parameter
        
        - Get request to retrieve metadata without passing survey_id
        - Assert status code is 400
        - Checks the error message in the response
        """

src/integration_tests/schemas/test_integration_unhappy_path.py
```diff
@@ -0,0 +1,170 @@
...
+    def test_get_schema_v2_404_not_found(self):
+        """
+        Test data not found by requesting a nonexistent GUID for GET /v2/schema.
+        * Expected: 404 Not Found.
+        """
+        response = self.session.get(
+            f"{config.API_URL}/v2/schema?guid=nonexistent_guid",
+            headers=self.headers,
+        )
+        assert response.status_code == 404
```
JacobHarrisonONS, line 170: could you add tests using nonsense in params e.g. /v2/schema?tfyghjkohjghfghgjhkjlghhj

src/integration_tests/schemas/test_integration_unhappy_path.py
```diff
@@ -0,0 +1,170 @@
...
+    pubsub_setup,
+    pubsub_teardown,
+    setup_session,
+    inject_wait_time,
+)
+from src.integration_tests.helpers.pubsub_helper import schema_pubsub_helper
+from src.test_data.schema_test_data import test_survey_id
+from src.test_data.shared_test_data import test_schema_subscriber_id
+
+class E2ESchemaIntegrationUnhappyPaths(TestCase):
```
JacobHarrisonONS, line 16: Could you move the tests into the schema happy path class and then name it something like SchemaEndpointsIntegrationTest please (if the class has not been named). the tests are not too complicated that they need to be moved. plus the docstrings make them quite easy to distinguish

JacobHarrisonONS (approved):

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -192,3 +195,231 @@ def test_survey_id_map(self):
...
+        response = self.session.post(
+            f"{config.API_URL}/v1/schema?survey_id={test_survey_id}",
+            json=invalid_data,
+            headers=self.headers,
+        )
+        assert response.status_code == 400
+        if "detail" in response.json():
+            assert response.json()["detail"] == "Invalid data format"
+        else:
+            print("Response JSON:", response.json())
```
IsaiahLai111, line 237: Conditional block should not exist in test assertion unless it is to act upon the configuration.
This block is not going to invalidate a response such as
`{"message": "Invalid data format"}` or `{"detail": "abc"}`
Because the `else` block allow the code to continue without any interruption
** Printing will have no effect when running test. If you need to log anything, use the logger

Change the conditional block into the following to properly invalidate a response:
`assert is_json_response(response)`
`assert "detail" in response.json()`
`assert response.json()["detail"] == "Invalid data format"`

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -192,3 +195,231 @@ def test_survey_id_map(self):
...
+        """
+        response = self.session.get(
+            f"{config.API_URL}/v1/schema?survey_id={invalid_survey_id}",
+            headers=self.headers,
+        )
+        assert response.status_code == 404
+        if "detail" in response.json():
+            assert response.json()["detail"] == "Schema not found"
+        else:
+            print("Response JSON:", response.json())
```
IsaiahLai111, line 255: Same as above

src/integration_tests/schemas/e2e_schema_integration_test.py
```diff
@@ -192,3 +195,231 @@ def test_survey_id_map(self):
...
+        """
+        response = self.session.get(
+            f"{config.API_URL}/v1/schema?survey_id={test_survey_id}",
+            headers=self.invalid_token_headers,
+        )
+        assert response.status_code == 401
+        if is_json_response(self, response) and "detail" in response.json():
+            assert response.json()["detail"] == "Unauthorized access"
+        else:
+            print("Non-JSON Response:", response.text)
```
IsaiahLai111, line 273: Same as mentioned, 401 do not require message validation

src/integration_tests/helpers/integration_helpers.py
```diff
@@ -339,3 +339,10 @@ def force_run_schedule_job():
         name=f"projects/{config.PROJECT_ID}/locations/europe-west2/jobs/trigger-new-dataset"
     )
     client.run_job(request=request)
+
+def is_json_response(self, response):
+    try:
+        response.json()
+        return True
+    except ValueError:
```
IsaiahLai111, line 347: `except Exception`

PR #231, opened 2026-08-10
Dataclass bug fix

angusgoody:
### Motivation and Context

If a field that is not in the UnitDataset dataclass is passed to this dataclass, the code breaks. This will occur if a legacy dataset is retrieved that contains a field that is deprecated. e.g `schema_version`

### What has changed

The code now filters out all fields that are not in the UnitDataset before creating the UnitDataset. 

### How to test?

Run the tests


jamesbryer (approved):
LGTM 🚀

ldearsley2 (approved):

